PR #24, opened 2018-12-12
Add proper support for unicode by encoding/decoding strings to/from UTF-8 when interacting with IDA.

### LINKED ISSUE #23: Fix for #20 results in unusable strings due to unicode strings not being handled by IDAPython.
This is fine and all that because everything is essentially being treated as ascii. This "was" the case, however, until the fix for issue #20 was requested. This fixed `db.get.string` so that it would properly decode the strings in the database instead of transforming them to UTF8 (IDA's solution). This fix resulted in the API always returning a unicode type.

However, unicode types aren't supported by `idaapi` which means that anything returned by `db.get.string` will need to be properly encoded before using it with IDAPython.

arizvisa:
IDA internally deals with unicode strings by encoding them to UTF-8. While this works, the transformation of the string actually changes its semantics. To fix this we literally need to fully unicode the entire repository. I'm not happy about this, but "it is what it is".

This adds a namespace to internal.interface named "string" which provides two methods. These are "to()" for converting a string into a format for IDA (encode to UTF-8), and "of()" which converts a string from IDA back into unicode (decode from UTF-8).

Closes #23


arizvisa:
This is such a messy fucking PR...but progress is happening at least.

Issue #29, opened 2019-11-09
Double-labeling issue when applying a name to an address belonging to a jumptable or a switch within a function


arizvisa:
When renaming a label for a switch (or a jumptable) with `database.name()`, both the specified name and the original name are displayed at the given address. This appears to be because both a global and local name are being applied to the same location.

The semantics for `minsc` warrant that all labels within a function will be converted to local whereas all names not within a function are to be converted to global global. Despite these constraints, the `listed` option still must be considered for both label types.

arizvisa:
Closed this fucker with PR #30.

PR #30, opened 2019-11-09
base/database: fixed an issue where local names would not override names with the global flag set.

arizvisa:
When applying a name to a label within a function with `database.name()`, in some instances more than one label was being displayed by IDA. It was determined that this is because in some disassembler modules IDA may apply the global flag to labels associated with switches or jump tables and that `database.name()` would not handle these cases.

This PR splits up the logic for `database.name()` into 3 distinctly separate closures. The first closure actually applies the name with whatever flags were determined, and the other two closures are used depending on whether a name is being applied to an address within a function or outside a function.

To fix the issue, `idaapi.get_visible_name()` is used to check if the visible name does not correspond to the name requested by the caller. If this is the case, then `idaapi.del_global_name()` is used to remove it. This results in all labels within a function having the `idaapi.SN_LOCAL` attribute set.

This closes issue #29.

### DIFF
--- a/base/database.py
+++ b/base/database.py
@@ -968,17 +968,59 @@ def name(ea, string, *suffix, **flags):
     res = (string,) + suffix
     string = interface.tuplename(*res)
 
-    # validate the address
+    # validate the address, and get the original flags
     ea = interface.address.inside(ea)
+    ofl = type.flags(ea)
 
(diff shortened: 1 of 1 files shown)


arizvisa:
Good job, Ali. You've been meaning to fix this shit for a while.

Issue #31, opened 2019-11-09
function.within() returns true on externals within the "extern" segment for elf files


arizvisa:
For some strange reason, `function.within()` is saying that externals in the "extern" segment is within a function. This appears to be IDA's doing because if you move your cursor to any of them, you can hit `Alt+P` and it'll let you edit its bounds as a function. Despite IDA being crazy here, `minsc` should detect this and return `false` since it's not actually a function.

arizvisa:
Thanks, Ilfak, for causing all of us users to have to hack around your hacks. \<grumbles\>

arizvisa:
Closed by PR #32.

PR #32, opened 2019-11-09
Fixed the issue where externals can be mis-identified as a valid function.

arizvisa:
In some IDBs that IDA makes for ELF files, the symbols inside the "extern" segment will be considered properly defined functions. This can result in `minsc` acting weird in those situations such as when applying/searching tags, or trying to determine how to name the given address.

This was fixed in `function.within()` by checking if the given address is within a segment with the type `idaapi.SEG_XTRN`. If so, then `function.within()` will properly return false. All of the tagging-related functions in the `database` module were also updated to double-check using `function.within()`.

The internal function, `interface.addressOfRuntimeOrStatic()`, needed a check as well. This function also verifies that the address is an actual import by cross-referencing the address with the imports list.

This closes issue #31.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -1092,6 +1092,19 @@ def addressOfRuntimeOrStatic(func):
         # yep, we're an import
         return True, func
 
+    # check if we're _not_ actually within a function (mis-defined external)
+    if not function.within(fn.startEA):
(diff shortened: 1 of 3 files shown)


arizvisa:
Whee...

Issue #36, opened 2019-11-25
Question: Which versions of IDA are supported?


karlek:
The documentation states that IDA versions: 6.9 up to 7.1.180227 are supported. 
https://arizvisa.github.io/ida-minsc/install.html#software-requirements

And the contribution guidelines states that new git tags will be added on new IDA versions. 
https://github.com/arizvisa/ida-minsc/blob/master/CONTRIBUTING.md#new-versions-of-ida-tags

As far as I can see, there are no git tags on this repo, and the project is actively developed. So I wonder if newer IDA versions are supported, like 7.4?

arizvisa:
Lol. Yeah... my bad. I haven't re-built the docs in a bit because nothing significant has really changed.

When I had tested 7.1.180227, all 3 platforms worked. All my friends however (whom I get most of my bug reports through) run it on Windows with 7.4, and they haven't pointed out anything. I can't be sure about the non-windows platforms though for 7.4 because we have all Windows licenses since that constraint.

I haven't upgraded my version of IDA since this was released because Ilfak's licensing has changed if one works for a company, which includes licensing server requirements and an explicit license per platform which makes it difficult for me to test this on absolutely everything.

Nonetheless, I \_will\_ still support this regardless. Probably till I die or stop reversing, whichever comes first.

arizvisa:
Oh also, this whole library is essentially just a compartmentalized wrapper around `idaapi`. So unless there are major changes to the API as there was from 6.95 to 7, we should still be okay unless something in `idaapi` is broken (IDA's tech support is phenomenal though when you point out things or need to understand internals). 

The only other requirement which I forgot to list is that it's developed with Python2, and I haven't gone through that Python3 transitioning because of my hate for it. I'll add this Python2 requirement to the docs and rebuild them as a result of this insight too.

arizvisa:
So, short answer: Yes. I will fix the bugz that you report in all versions of IDA from 5.6 and up.

arizvisa:
This issue was addressed by PRs #39, and #40. The documentation in the `gh-pages` branch was also re-built to accommodate the addressed changes so that https://arizvisa.github.io/ida-minsc

@karlek, do these changes address your question/concerns?

karlek:
> Lol. Yeah... my bad. I haven't re-built the docs in a bit because nothing significant has really changed.

[...]

PR #39, opened 2019-11-27
Updated CONTRIBUTING.md to remove policies that just haven't been honored

arizvisa:
It was pointed out in issue #36 that a number of things haven't been honored during the present lifetime of this project. This PR updates a number of things within the documentation that just aren't relevant anymore as a result.

The documentation is also updated to reference that later versions of IDA are supported and also that at the current moment only Python2 is supported despite this likely changing in the future.

This aims to close the concerns posed by issue #36.


arizvisa:
This only closes part of the issue in #36 wrt CONTRIBUTING.md mentioning the whole tagging thing. I'm the only developer, so if more people end up using minsc then the removed section can be re-introduced.

PR #40, opened 2019-11-27
Updated documentation to reference latest version of IDA and the Python 2.x requirement for IDAPython

arizvisa:
The documentation references an older version of IDA which can mislead users into believing that the latest version of IDA is unsupported. This is not the case and as such this PR updates the documentation to reference the present version of IDA.

The index and installation pages have also been updated to reference the one hard requirement which is Python 2.x. This is likely to change in the future, but at the present time this should be noted as users are likely migrating to 3.x due to the EOL of the 2.x series.

This is part of PR #39, and aims to fix the concerns posted by issue #36.


arizvisa:
This merge will require a rebuild of the documentation.

Issue #53, opened 2020-06-25
AArch and AArch64 register support (KeyError raised)?


arizvisa:
It's been pointed out that AArch64 register support is lacking.  When accessing a register off of the architecture's dictionary a misleading KeyError will be raised. This is actualy due to the AArch64 registers not being completely defined. This includes the `X*`, the `W*` registers, and possibly a few more. It was mentioned that there might be a few registers for the (32-bit) AArch architecture that are not defined as well.

Please add these as the architectures are somewhat useless without them.

arizvisa:
Closed by PR #54.

PR #54, opened 2020-06-25
Add the rest of the missing registers for both AArch32 and AArch64 support.

arizvisa:
Some of the registers for the AArch64 architectures were not defined whilst only supporting the 32-bit registers and ignoring the relationships between them. This PR adds the rest of them and includes relationships for all the registers and flags that IDA knows about (and a few more). There was also a bug that needed to be fixed due to the `idaapi.dt_bitfield` type actually being (mis-)named as `idaapi.bitfild`.

While testing this I also noticed that some arm instructions have an operand type of 14 (0xe). Unfortunately this operand type isn't listed in IDA as the `idaapi.o_*` properties only go up to 13 (0xd). I added an `unknown` operand type that catches this and warns the user about this. The semantics of operand type 14 will need eventually be reversed out of IDA at some point...but not today.

This fixes issue #53.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -1566,12 +1566,14 @@ class architecture_t(object):
         # older
         if idaapi.__version__ < 7.0:
             dtype_by_size = internal.utils.fcompose(idaapi.get_dtyp_by_size, six.byte2int)
+            dt_bitfield = idaapi.dt_bitfield
         # newer
         else:
             dtype_by_size = idaapi.get_dtype_by_size
+            dt_bitfield = idaapi.dt_bitfild
 
         #dtyp = kwargs.get('dtyp', idaapi.dt_bitfild if bits == 1 else dtype_by_size(bits//8))
-        dtype = six.next((kwargs[n] for n in ('dtyp', 'dtype', 'type') if n in kwargs), idaapi.dt_bitfield if bits == 1 else dtype_by_size(bits // 8))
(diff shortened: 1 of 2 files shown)


arizvisa:
This PR is waiting on an associate to test that it works for him. In the meantime I'm running `ins.ops()` on all my AArch32 and AArch64 samples to try and squash any other bugs.

arizvisa:
Force-pushed to fix some incorrectly worded commits.

arizvisa:
Some of the types for the sub-registers actually needed to be floating point, so that was corrected.

arizvisa:
It turns out that IDA uses different registers depending on whether the 32-bit or 64-bit flavor of arm is being used. I included an additional that handles this.

arizvisa:
On AArch32, there's a different `dt_type` that's used on a register when doing sign-extensions. We don't really support that, so I suffixed the relevant registers with their size (similar to AT&T) so that the register types can be distinguished by name.

arizvisa:
It turns out that there's two operand types that IDA hasn't documented. One of them is `idaapi.o_idpspec4`, which seems to be an extension register list. This is likely organized in some sort of bitflag oriented structure of which I'm too lazy to figure out arm encoding again in order to determine. The other type doesn't have an idpspec, and is assigned as 14. This seems to be used for conditionals, but only seems to use one attribute that I think is an enumeration of some sort.

Issue #55, opened 2020-06-26
MIPS64 architecture support (KeyError and other exceptions raised when decoding an operand)


arizvisa:
It was pointed out that only support for the MIPS32 architecture is implemented. When accessing an operand off of the arch MIPS64 architecture's dictionary the value returned from `ins.op_bits()` does not correlate to the actual operands. This is actually due to MIPS64 architecture not actualy having been implemented yet.

I guess it's now time to do this... :-/

arizvisa:
Turns out that the reason why these were missed are because IDA's register list doesn't always include all the available registers for the architecture. So, the a4-a8 registers were missed as well as a couple of others.

arizvisa:
This is being worked on at #56, and will eventually be closed by it.

arizvisa:
Closed by PR #56.

PR #56, opened 2020-06-27
Split up the MIPS architecture into its 32-bit and 64-bit variants, and added missing registers or operand decoders to it.

arizvisa:
The MIPS architecture implementation was missing some of the registers as it turns out that IDA doesn't include all available registers in the database register names which the definition was actually based on. It also turns out that the operand decoders were wrong as `idaapi.o_idpspec0` represents trap codes which can be encoded in the `teq` instruction, and `idaapi.o_idpspec1` represents a floating-point register by index. This is not to mention that the bit sizes of each operand also differed between 32-bit and 64-bit versions despite them not really affecting anything.

This PR splits up the MIPS architecture into a 32-bit and 64-bit version (similar to the ARM architectures), and corrects the sizes used for the registers. A couple more coprocessor registers were also added and the tables were cleaned up so that they're aligned and easier to read. The logic for searching for the different register types was also refactored so that they're defined as methods that can be called from the `architecture_t`. After this was done, the operand decoders were fixed up so that the trap code is in its own tuple, and the floating-point register operand-type decoder was introduced.

This closes issue #55.

### DIFF
--- a/base/instruction.py
+++ b/base/instruction.py
@@ -61,7 +61,7 @@ class __optype__(object):
     cache = {}
     @classmethod
     def define(cls, processor, type):
-        '''Register the operand decoder for the specfied `processor` an `type`'''
+        '''Register the operand decoder for the specfied `processor` and `type`'''
         def decorator(fn):
             res = processor, type
             return cls.cache.setdefault(res, fn)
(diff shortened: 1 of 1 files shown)


base/instruction.py
```diff
@@ -1783,24 +1794,44 @@ def symbols(self):
             r, _ = self
             yield r
 
+    class trap(interface.namedtypedtuple, interface.symbol_t):
+        """
+        A tuple for representing a trap code that can be encoded within
+        certain instructions on the MIPS architecture.
+
+        Simply wraps the encoded integer as a individual value in a tuple
```
arizvisa, line 1802: should be "an" instead of "a".

base/instruction.py
```diff
@@ -1783,24 +1794,44 @@ def symbols(self):
...
+        Takes a `regnum` argument which returns the correct register.
+        """
+        global architecture
+        return architecture.by_coprocessor(regnum)
+
     @staticmethod
-    def coproc(regnum):
+    def float(regnum):
         """
         A callable that returns a co-processor for the MIPS architecture.
```
arizvisa, line 1829: The documentation for this could reference the coprocessor index. Also remove the "-" (hyphen) when describing it.

arizvisa:
Okay. Merging this.

Issue #57, opened 2020-07-08
The db.get.string() can not run at ida 6.8


gool123456:
Hi @arizvisa , this method Show Error in ida 6.8 :

`Python>db.get.string(0x00000001409D9B70)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 290, in database.string
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 4877, in string
    strtype = idaapi.get_str_type(ea)
AttributeError: 'module' object has no attribute 'get_str_type'`

--------------------------------------------------------------------------

The code I modified is passed, can you pull it into the project ?  :
```

   @classmethod
    def string(cls, ea, **length):
        """Return the array at the address specified by `ea` as a string.

        If the integer `length` is defined, then use it as the length of the array.
        """

        # Fetch the string type at the given address
        if idaapi.__version__ < 7.0:
            strtype = idc.GetStringType(ea)
        else: 
            strtype = idaapi.get_str_type(ea)

        # If no string was found, then try to treat it as a plain old array
        # XXX: idaapi.get_str_type() seems to return 0xffffffff on failure instead of idaapi.BADADDR
        if strtype in {idaapi.BADADDR, 0xffffffff}:
            res = cls.array(ea, **length)

            # It wasn't an array and was probably a structure, so we'll just complain to the user about it
            if not isinstance(res, _array.array):
                raise E.InvalidTypeOrValueError(u"{:s}.string({:#x}{:s}) : The type at address {:#x} cannot be treated as an unformatted array and as such is not convertible to a string.".format('.'.join((__name__, cls.__name__)), ea, u", {:s}".format(utils.string.kwargs(length)) if length else '', ea))

            # Warn the user and convert it into a string
            logging.warn(u"{:s}.string({:#x}{:s}) : Unable to automatically determine the string type at address {:#x}. Treating as an unformatted array instead.".format('.'.join((__name__, cls.__name__)), ea, u", {:s}".format(utils.string.kwargs(length)) if length else '', ea))
            return res.tostring()

        # Get the string encoding (not used)
        encoding = idaapi.get_str_encoding_idx(strtype)

        # Get the terminal characters that can terminate the string
[...]

PR #58, opened 2020-07-08
Fixed a compability issue with the `database.get.string` function on earlier versions of IDA.

arizvisa:
@gool12346 reported in issue #57 that the `database.get.string()` function doesn't work in IDA 6.8. This is due to the non-existence of the `idaapi.get_str_type`. Earlier versions of IDA actually store the string type information in the `idaapi.opinfo_t` for the given address. This API change is actually laid out in the IDA porting guide at https://www.hex-rays.com/products/ida/support/ida74_idapython_no_bc695_porting_guide.shtml.

A patch was provided by the original reporter, which had fixed the compatibility issue for the earlier versions, but mistakenly broke it in the later versions. His patch was refactored to not use the `idc` module instead opting for a purely `idaapi`-based implementation, and then the calls to `idaapi.get_str_type_code` were replaced with a composition that automatically returns an integer if an earlier version is detected.

This should close issue #57.

### DIFF
--- a/base/database.py
+++ b/base/database.py
@@ -5467,7 +5467,10 @@ class get(object):
         """
 
         # Fetch the string type at the given address
-        strtype = idaapi.get_str_type(ea)
+        if idaapi.__version__ < 7.0:
+            strtype = idc.GetStringType(ea)
+        else:
+            strtype = idaapi.get_str_type(ea)
 
         # If no string was found, then try to treat it as a plain old array
         # XXX: idaapi.get_str_type() seems to return 0xffffffff on failure instead of idaapi.BADADDR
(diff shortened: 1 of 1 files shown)


gool123456:
@arizvisa Hi, Before ida-minsc is officially compatible with versions prior to 6.8, I tried to modify the interface of ida-minsc and named it myIdaOsc in the site-packages directory. I can use it by importing myIdaOs. I also modified db.get .string(), I also forgot to modify the specifics...just simply added the non-existing variables from the idaapi.py declaration of 7.2 to the declaration of 6.8 idaapi.py...

gool123456:
@arizvisa Great! I use GH-57 and everything works smoothly. Thank you ^^

arizvisa:
@gool123456, what non-existing variables did you need to add? Do I need to fix the usage of them in this PR?

gool123456:
@arizvisa No need to fix PR, I use GH-57, it is normal. I said myIdaOs is the old version of ida-minsc, at that time was not compatible with ida6.8, I can only do this.

arizvisa:
Okay. Thanks. I'm going to merge this PR then.

Issue #62, opened 2020-07-27
Unable to set a structure member's custom type, it always is 'int'


arizvisa:
@arizvisa hi, I have to come and ask you questions, I don't know what went wrong, I can't set the member's custom type , it awaly is 'int':

```
import idaapi
import idc


st1 =structure.by_name('TTDBTextInfoGroup')
struct_name = 'TTDBTextInfoGroup*'
f1 = st1.members.by(offset=0)

print f1.typeinfo.dstr()
ordinal = f1.typeinfo.get_ordinal()
f1.typeinfo.set_numbered_type(idaapi.cvar.idati,ordinal,BTF_STRUCT, struct_name)
print f1.typeinfo.dstr()
```

_Originally posted by @gool123456 in https://github.com/arizvisa/ida-minsc/issues/61#issuecomment-664488530_

arizvisa:
Can you print out the structure st1 to me? There might not be enough space in the field to promote it to an int.

This is actually related to issue #45. IDA doesn't return an error or anything, so to solve issue #45, we need to verify that the field was properly assigned.

_Originally posted by @arizvisa in https://github.com/arizvisa/ida-minsc/issues/61#issuecomment-664504271_

arizvisa:
Also, are you trying to change the type of the member at offset 0 to a structure?

If so, then you'll want to do:
```
st = struc.by('mystructure')
m = st.members.by(offset=0)
m.type = struc.by('structuremember')
```

_Originally posted by @arizvisa in https://github.com/arizvisa/ida-minsc/issues/61#issuecomment-664505803_

arizvisa:
You can also treat `member.typeinfo` as a string, meaning you can do:
```
m.typeinfo = 'char*'
```

_Originally posted by @arizvisa in https://github.com/arizvisa/ida-minsc/issues/61#issuecomment-664506222_

gool123456:
@ Thanks , but ida 6.8 unfind the idaapi.get_idati , The error happend at structure.py -> typeinfo line 1619 :

I try fix it :

@typeinfo.setter
@utils.string.decorate_arguments('info')
def typeinfo(self, info):
    '''Set the type info of the member to `info`.'''
    if idaapi.__version__ < 7.0:
        **til, ti = idaapi.cvar.idati , idaapi.tinfo_t()   #added**
    else:
        til, ti = idaapi.get_idati(), idaapi.tinfo_t(),



But then the error occurred :

```
[...]

PR #63, opened 2020-07-28
Fixed some compatibility issues with the usage of type libraries and declaration parsing

arizvisa:
Earlier versions of IDA do not support the `idaapi.get_idati()` function for determining the default type library as well as the prototype for the `idaapi.parse_decl` function which is also different.

This PR fixes these issues by replacing all instances of `idaapi.get_idati()` on earlier versions of IDA with the `idaapi.cvar.idati` property (@gool123456's solution). All instances of `idaapi.parse_decl` are also now calling the older `idaapi.parse_decl2` function with the correct prototype.

This should close the issue posed in #62.

### DIFF
--- a/base/_declaration.py
+++ b/base/_declaration.py
@@ -77,7 +77,9 @@ def parse(info):
     # that we're responsible for raising an exception if there's a parsing
     # error of some sort. If it succeeds, then we can return our typeinfo.
     # Otherwise we return None because of the inability to parse it.
-    if idaapi.__version__ < 7.0:
+    if idaapi.__version__ < 6.9:
+        return None if idaapi.parse_decl2(til, terminated, None, ti, idaapi.PT_SIL) is None else ti
+    elif idaapi.__version__ < 7.0:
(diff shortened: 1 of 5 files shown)


arizvisa:
@gool123456, can you check this PR and confirm that it fixes issue #62? You should be able to see the local-types when you view structures with `struc.list()`, you should also see members like `struc.by('GUID').members`. Then assigning types, you should be able to do it with both `member.typeinfo` and `member.type` like:

```
st = func.frame()
print(st.members)

st = struc.new('mystructure')
st.members.add('myfield', (int,2))

print(st.members[0].size)
print(st.members[0].type)

st.members.by('myfield').type = [(int,1), 2]
print(st.members.by('myfield').typeinfo)

st.members.by('myfield').typeinfo = 'unsigned short'
print(st.members.by('myfield'))
```

To try it out you can use `git` like the following which will check out the PR into its own branch.
```
$ git fetch origin pull/63/head:fix-issue-62
```

gool123456:
@arizvisa Your example is ok ! This RP structure has been repaired.

arizvisa:
Awesome. Thanks you.

Merging this PR.

Issue #65, opened 2020-07-30
The netnode for "$ tagcache" seems to be being reused between separate databases.


arizvisa:
The "$ tagcache" capability of ida-minsc is based around the contents of a blob started in a netnode. The idea is that the netnode is created when the database is created and then just referenced everytime it is needed. It seems that the logic for creating the netnode's identifier is scoped according to the `internal.comment.tagging` namespace as opposed to being scope according to the database itself. This netnode is instantiated by either the "idp.ev_init" or "idp.init" hooks.

The issue is because the identifier for this netnode is started within the class' definition and then only updated if the attribute is missing. This means that we're essentially trashing any blob that has the same netnode id when we open up another database which can also result in some weird values being displayed within the cache.

The entirety of the busted logic is in the implementation of the `tagging` base class within `base/_comment.py`:
```
    @classmethod
    def __init_tagcache__(cls, idp_modname):
        cls.node()
        logging.fatal(u"{:s}.init_tagcache('{:s}') : Initialized tagcache with netnode \"{:s}\" and node id {:#x}.".format('.'.join(('internal', __name__, cls.__name__)), internal.utils.string.escape(idp_modname, '\''), internal.utils.string.escape(cls.__node__, '"'), cls.__nodeid__))

    @classmethod
    def __nw_init_tagcache__(cls, nw_code, is_old_database):
        idp_modname = idaapi.get_idp_name()
        return cls.__init_tagcache__(idp_modname)

    @classmethod
    def node(cls):
        if hasattr(cls, '__nodeid__'):
            return cls.__nodeid__
        node = internal.netnode.get(cls.__node__)
        if node == idaapi.BADADDR:
            node = internal.netnode.new(cls.__node__)
        cls.__nodeid__ = node
        return node
```

arizvisa:
This issue is fixed by PR #66.

arizvisa:
Closing this issue as it was fixed by PR #66.

PR #66, opened 2020-07-30
Fixed a logic issue in the `tagging.node()` function which would mistakenly scope the netnode id with the class instead of with the database.

arizvisa:
This PR fixes some busted logic in the `internal.comment` module which would result in the netnode id that was allocated for the tag-cache being re-used between databases. This was because the id for the netnode was being stored within the class, and the logic for executing the `tagging.__init_tagcache__` hook was not properly ensuring that the cached identifier was being updated.

This would result in the weird symptom of there occasionally being bunk-values within the tag-cache when opening up other databases within the same instance of IDA. In all actuality, this was thrashing the values in the netnode that was cached which would've likely broken something in a database but by sheer luck hasn't done that yet. 

Rather than deleting the cached id everytime the hook is run, a new parameter was introduced to the `tagging.node()` function that would specify to use the cached version by default. If the flag in the parameter is changed, then the node id will be re-fetched or created depending on what's necessary. After introducing this new parameter to `tagging.node()`, the hook's usage of the function was modified to \_always\_ fetch (or create) the most recent id for the netnode.

This closes issue #65.

### DIFF
--- a/base/_comment.py
+++ b/base/_comment.py
@@ -642,8 +642,9 @@ class tagging(object):
 
     @classmethod
     def __init_tagcache__(cls, idp_modname):
-        cls.node()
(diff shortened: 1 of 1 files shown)


arizvisa:
Merging this PR as it seems to fix the related.

Issue #61, opened 2020-07-23
ReadorWriteError exception raised when trying to use comment api in IDA 6.8


arizvisa:
This issue manifested after fixing issue #59 with PR #60.

```
Python>db.comment(0x006DA9CF, "123")
CRITICAL:root:internal.interface.priorityhook.callback(7186895, False) : Callback for "IDB_Hooks.cmt_changed" with priority (+0) raised an exception while executing <bound method type.old_changed of <class 'hooks.address'>>
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : Traceback ("IDB_Hooks.cmt_changed" was hooked at)
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\misc\hooks.py", line 1059, in ida_is_busy_sucking_cocks
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : make_ida_not_suck_cocks(idaapi.NW_INITIDA)
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\misc\hooks.py", line 993, in make_ida_not_suck_cocks
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : ui.hook.idb.add('cmt_changed', address.old_changed, 0)
WARNING:root:internal.interface.priorityhook.callback(7186895, False) :
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : Traceback (most recent call last):
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base_interface.py", line 373, in closure
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : result = callable(*parameters)
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\misc\hooks.py", line 219, in old_changed
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : internal.comment.contents.set_address(ea, 0)
WARNING:root:internal.interface.priorityhook.callback(7186895, False) : File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base_comment.py", line 971, in set_address
[...]

PR #67, opened 2020-07-30
Fixed a couple of compatibility issues with hooks that are being executed in response to IDA's auto-analysis.

arizvisa:
When IDA runs its auto-analysis, a number of hooks are executed in order to ensure that the tag-cache is updated in response to IDA's changes. When executing these hooks (specifically when earlier versions of IDA apply a type library to the database), a handful of unhandled exceptions are raised which are pretty alarming. The hooks that were causing these issues were the `idb.func_tail_removed`, and the `idb.cmt_changed` hooks.

Specifically there were 3 types of exceptions that were being raised by both of these two hooks.
- ReadOrWriteError -- which claimed that it was unable to write to the cache at a given address
- ValueError -- which was because the `min()` and `max()` keywords were being used on an empty list
- Some kind of null pointer exception from swig -- which was because of a comparison of an `idaapi.area_t` to `None`.

After cleaning up some of the exception handling and adding logs to `internal.comment`, it was discovered that the first `ReadOrWriteError` was happening because an empty value was being passed to the `contents._write` function, and the result of an API call was being blindly returned. This resulted in the function trying to remove a netnode from the given address, and since there wasn't anything at that address, the value `0` was returned. As this value is an implicit false, a sanity check was raising an exception thinking some kind of critical error was happening which was not the case.

The second `ValueError` exception was being raised within the `idb.func_tail_removed` hook. This issue was happening because the implementation of the hook was filtering out addresses that were outside of the function in order to determine where to move any tags that are found. If the filter ended up not finding any addresses, the resulting list would be empty. Both the `min()` and `max()` keywords raise an exception on an empty list, which is where this one came from.

The last exception was very strange in that the `idb.func_tail_removed` hook was performing the aforementioned filter to determine what addresses were within a function versus being outside a function. This filter is being done with the `idaapi.get_func(ea)` api which should return `None` when the given address is not within a function. For some strange reason, comparing an `idaapi.area_t` with `None` will result in a null-pointer exception being raised by swig from the `area_t.__ne__` method on earlier versions of IDA. This doesn't happen in all cases, and I couldn't reproduce it with just any old `idaapi.area_t` or `idaapi.func_t`. It seems that when being called from the `idb.func_tail_removed` hook, its `pfn` parameter is in some weird state where when it is compared to `None`, its SWIG wrapper tries to dereference it.

The first exception was fixed by simply treating the count that was being returned as a non-critical issue. The second exception was fixed by checking if the list was filtered to empty, and then terminating instead of attempting to process it. The last exception from SWIG was fixed by explicitly checking if the address is going to be `None` prior to comparing it against the `pfn` parameter. This way we can avoid comparing it against `None` and excluding the item before letting SWIG do its comparison.

This should hopefully close issue #61.

### DIFF
--- a/base/_comment.py
+++ b/base/_comment.py
@@ -714,7 +714,7 @@ class contents(tagging):
         try:
             data, sz = cls.codec.decode(encdata)
             if len(encdata) != sz:
-                raise internal.exceptions.SizeMismatchError(u"{:s}._read_header({!r}, {:#x}) : The number of bytes that was decoded ({:+#x}) did not match the expected size ({:+#x}).".format('.'.join(('internal', __name__, cls.__name__)), target, ea, sz, len(encdata)))
+                raise internal.exceptions.SizeMismatchError(u"{:s}._read_header({!r}, {:#x}) : The number of bytes that was decoded did not match the expected size ({:#x}<>{:#x}).".format('.'.join(('internal', __name__, cls.__name__)), target, ea, sz, len(encdata)))
         except Exception as E:
             logging.warn(u"{:s}._read_header({!r}, {:#x}) : An exception ({!s}) was raised while trying to decode the header for address {:#x} from sup cache associated with key {:#x}.".format('.'.join(('internal', __name__, cls.__name__)), target, ea, E, ea, key), exc_info=True)
             logging.info(u"{:s}._read_header({!r}, {:#x}) : Failure decoding the following data from sup cache: {!r}.".format('.'.join(('internal', __name__, cls.__name__)), target, ea, encdata))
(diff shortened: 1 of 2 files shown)


arizvisa:
As the issue that this PR fixes was only reproduceable for me when opening up a second database within the same instance of IDA, I suspect this issue is also related to #65, and thus PR #66 might also influence the bugs fixed by this PR (and mentioned in #61).

arizvisa:
This was confirmed by the OP. Just need a review.

base/_comment.py
```diff
@@ -805,8 +808,13 @@ def _write(cls, target, ea, value):
                 ok = cls._write_header(target, ea, None)
                 if not ok:
                     logging.debug(u"{:s}._write({!r}, {:#x}, {!s}) : Unable to remove address from sup cache with the key {:#x}.".format('.'.join(('internal', __name__, cls.__name__)), target, ea, internal.utils.string.repr(value), key))
+
             finally:
-                return internal.netnode.blob.remove(key, cls.btag)
+                count = internal.netnode.blob.remove(key, cls.btag)
+                if not count:
+                    logging.debug(u"{:s}._write({!r}, {:#x}, {!s}) : Ignoring failed removal of non-existent blob ({!s}) for the key {:#x}.".format('.'.join(('internal', __name__, cls.__name__)), target, ea, internal.utils.string.repr(value), cls.btag, key))
+
```
arizvisa, line 816: It'd be good to just add a log anyways that displays how many blobs were removed from the specified address.

base/_comment.py
```diff
@@ -851,7 +861,7 @@ def iterate(cls):
             encdata = internal.netnode.sup.get(node, ea)
             data, sz = cls.codec.decode(encdata)
             if len(encdata) != sz:
-                logging.warn(u"{:s}.iterate() : Failed decoding tag names out of sup cache for {:#x} due to the length of encoded data ({:#x}) not matching the expected size ({:#x}).".format('.'.join(('internal', __name__, cls.__name__)), ea, len(encdata), sz))
+                logging.warn(u"{:s}.iterate() : Failed decoding tag names out of sup cache for {:#x} due to the length of encoded data ({:+#x}) not matching the expected size ({:#x}).".format('.'.join(('internal', __name__, cls.__name__)), ea, len(encdata), sz))
```
arizvisa, line 864: add the word "address" here when display what sup cache is being referenced.

PR #70, opened 2020-08-13
Created a filesystem abstraction around IDA's netnode API to allow interacting with large files in the database.

arizvisa:
This PR wraps the IDAPython netnode API (supvals, and blobs) in order to allow the storage of arbitrarily-named and arbitrarily-sized files in a database. This is done using two types of netnodes. The `$ filesystem.index` netnode contains the directory index which is flat and is treated pretty much like a keystore, and then the `$ filesystem.contents` netnode which is responsible for containing the actual contents of the files described in the directory index.

Normally one could use the netnode API and simply store their contents in a blob, but due to the way IDAPython exposes this API the entire contents of the blob needs to be completely loaded into memory when reading or writing. The filesystem abstraction implemented by this PR thus allows a user to storage arbitrarily-large files and read them as an i/o stream instead of having to load the entire contents of the file in memory.

This feature was suggested in issue #8 and aims to close it.

### DIFF
--- a/base/_filesystem.py
+++ b/base/_filesystem.py
@@ -1,0 +1,618 @@
+import six, functools, operator, logging, codecs, array, collections
+import internal, idaapi
+
+SECTOR = 1024
+INTEGER_SIZE = 8
+
+class namedtypedtuple(tuple):
+    """
+    A named tuple with actual type checking.
+    """
+    _fields = ()
+    _types = ()
+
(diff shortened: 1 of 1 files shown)


arizvisa:
Note to self: make sure the filesystem protocol is versioned as currently it's using an allocation table, and well...defragmenting sucks.

Issue #71, opened 2020-10-08
The tag cache can get out of sync when writing comments of the wrong type for a given address


arizvisa:
When updating tags, the values in the tagging netnode will occasionally desynchronize. Although this eventually gets fixed the next time a comment is made at the miscached address, this can result in queries fetching an item from the cache that was already removed. This issue is hidden because the results from both `database.tag`, and `function.tag` are checked for the keys that the user requests prior to yielding their result, but this check shouldn't be necessary anyways.

arizvisa:
After adding some debug logging, it turns out that the cache was getting desynchronized because I had used the wrong comment type and just didn't notice it.

arizvisa:
This issue is being worked on in PR #72.

arizvisa:
This was closed by PR #72

PR #72, opened 2020-10-08
Modified the support for tags so that they can be encoded in both repeatable and non-repeatable comments.

arizvisa:
This PR changes the way that tags work so that they can be included in either repeatable or non-repeatable comments without issue. Prior to this, we were forcing which type the user was using when defining their comment (and hence their tags). This was being done in two ways. The first way was to check which comment was being used during the encoding of a tag, and if the wrong type was being used then the logic would be to move the contents from the wrong comment location to the correct one. The second way this was forced was by hooking when comments were being defined by the user, and then manually rewriting them to the correct comment location. This meant that the tag cache would keep track of only the tags that were encoded to the right place which resulted in the cache becoming desynchronized until the user modified the comment again.

The PR remedies the aforementioned logic by trying to determine which comment type the user's tag was written to. If the tag was not found in either comment location, then we explicitly choose the correct one that makes sense for the given address. When tags are fetched they are still combined like before, but the tags from the correct comment type are now given priority. Also, when writing or erasing tags we're now guarding against event hooks being triggered to prevent the references in the cache being tampered with more than once.

This PR also changes the way the tag cache works in that now it keeps track of tags that have been encoded in both types of comments at a given address. This means that if a user has more than one comment type defined, then the name cache will include the count for all tag names with duplicates, and the address cache will be the sum of all of the tags that have been defined at both locations.

This closes issue #71

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -286,6 +286,10 @@ class prioritybase(object):
             continue
         return not notok
 
+    @property
+    def available(self):
+        return {item for item in self.__cache__}
+
     def enable(self, target):
         '''Enable any callables for the specified `target` that has been previously disabled.'''
         if target not in self.__disabled:
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -288,6 +288,7 @@ class prioritybase(object):
 
     @property
     def available(self):
+        '''Return all of the available targets that can be either enabled or disabled.'''
         return {item for item in self.__cache__}
 
     def enable(self, target):
--- a/base/database.py
+++ b/base/database.py
@@ -1537,9 +1537,12 @@ def tag(ea, key, value):
         func = None
     repeatable = False if func and function.within(ea) else True
 
-    # grab the current tag out of the correct repeatable or non-repeatable comment
+    # figure out which comment the tag is in, if it's in neither then use the
+    # comment type that's the correct one.
     ea = interface.address.inside(ea)
-    state = internal.comment.decode(comment(ea, repeatable=repeatable))
+    state_correct = internal.comment.decode(comment(ea, repeatable=repeatable))
+    state_wrong = internal.comment.decode(comment(ea, repeatable=not repeatable))
+    state, where = (state_correct, repeatable) if key in state_correct else (state_wrong, not repeatable) if key in state_wrong else (state_correct, repeatable)
 
     # update the tag's reference if we're actually adding a key and not overwriting it
     if key not in state:
@@ -1559,7 +1562,7 @@ def tag(ea, key, value):
     except Exception:
         raise
     else:
-        comment(ea, internal.comment.encode(state), repeatable=repeatable)
+        comment(ea, internal.comment.encode(state), repeatable=where)
     finally:
         [ ui.hook.idb.enable(item) for item in hooks ]
 
@@ -1594,7 +1597,10 @@ def tag(ea, key, none):
 
     # fetch the dictionary from the other repeatable/non-repeatable comment
     # and set a flag in case we need to clear it out for being in the wrong place
-    state = internal.comment.decode(comment(ea, repeatable=repeatable))
+    state_correct = internal.comment.decode(comment(ea, repeatable=repeatable))
+    state_wrong = internal.comment.decode(comment(ea, repeatable=not repeatable))
+    state, where = (state_correct, repeatable) if key in state_correct else (state_wrong, not repeatable) if key in state_wrong else (state_correct, repeatable)
+
     if key not in state:
         raise E.MissingTagError(u"{:s}.tag({:#x}, {!r}, {!s}) : Unable to remove tag \"{:s}\" from address.".format(__name__, ea, key, none, utils.string.escape(key, '"')))
     res = state.pop(key)
@@ -1607,7 +1613,7 @@ def tag(ea, key, none):
     except Exception:
         raise
     else:
-        comment(ea, internal.comment.encode(state), repeatable=repeatable)
+        comment(ea, internal.comment.encode(state), repeatable=where)
     finally:
         [ ui.hook.idb.enable(item) for item in hooks ]
 
--- a/base/database.py
+++ b/base/database.py
@@ -1444,26 +1444,36 @@ def tag(ea):
     ea = interface.address.inside(ea)
 
     # if not within a function, then use a repeatable comment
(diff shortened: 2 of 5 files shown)


base/_interface.py
```diff
@@ -286,6 +286,10 @@ def unhook(self):
             continue
         return not notok
 
+    @property
+    def available(self):
+        return {item for item in self.__cache__}
```
arizvisa, line 291: needs documentation.

base/database.py
```diff
@@ -1550,9 +1551,22 @@ def tag(ea, key, value):
         else:
             internal.comment.globals.inc(ea, key)
 
-    # now we can actually update the tag and encode it into the comment
+    # grab the previous value, and update the state with the new one
```
arizvisa, line 1554: this comment does not correspond to the logic.

base/database.py
```diff
@@ -1581,14 +1595,27 @@ def tag(ea, key, none):
         func = None
     repeatable = False if func and function.within(ea) else True
 
-    # fetch the dict, remove the key, then write it back.
-    state = internal.comment.decode(comment(ea, repeatable=not repeatable))
-    state and comment(ea, '', repeatable=not repeatable) # clear the old one
-    state.update(internal.comment.decode(comment(ea, repeatable=repeatable)))
+    # fetch the dictionary from the other repeatable/non-repeatable comment
+    # and set a flag in case we need to clear it out for being in the wrong place
```
arizvisa, line 1599: this comment does not at all describe what's happening.

base/function.py
```diff
@@ -1640,10 +1640,26 @@ def tag(func, key, value):
     if key == '__typeinfo__':
         return type(fn, value)
 
-    # decode the comment, fetch the old key, re-assign the new key, and then re-encode it
-    state = internal.comment.decode(comment(fn, repeatable=True))
+    # decode both comments and figure out which type of comment the tag is
+    # currently in. if it's in  neither then we just fall back to a repeatable
```
arizvisa, line 1644: there's a stray space here.

custom/tagfix.py
```diff
@@ -40,47 +40,19 @@ def fetch_contents(fn):
...
+        for items in map(internal.comment.decode, [repeatable, nonrepeatable]):
+            #items.pop('name', None)
+            for name in items:
+                addr[ea] = addr.get(ea, 0) + 1
+                tags[name] = tags.get(name, 0) + 1
+            continue
         continue
     return func.address(fn), addr, tags
 
-def check_contents(ea):
```
arizvisa, line 51: this removal will require a documentation update.

arizvisa (commented):
don't forget to test the `custom.tagfix.everything()` function.

arizvisa (commented):
Didn't see anything else. Time to test!

arizvisa:
Running the `custom.tagfix.everything()` function overnight seemed to work.

arizvisa:
Both repeatable and non-repeatable tags for contents seem to be being tallied properly, and their priority seems to be correct.

arizvisa:
Seems like global tags work the same way...however, there's a discrepancy when making a defining a regular comment against a runtime-linked address.

So apparently when you hit ";" over a runtime-linked address, IDA defines this as a repeatable "function" comment. When you hit "Shift+;" over a runtime-linked address, IDA defines this as a non-repeatable "database" comment.

This means that when `function.tag` hands over its logic to `database.tag`, the `database.tag` function is using the wrong style of comment and should be reading with `function.comment` instead of `database.comment`. Since this is a runtime-linked function and shouldn't be in a function at all, we need to ensure that we read/write runtime-linked comments with `function.comment` when using `database.tag`.

/me sighs

base/database.py
```diff
@@ -1577,21 +1615,43 @@ def tag(ea, key, none):
...
+        rt, func = False, None
+    repeatable = False if func and function.within(ea) and not rt else True
+
+    # figure out which comment type the user's key is in so that we can remove
+    # that one. if we're a runtime-linked address, then we need to remove the
+    # tag from a repeatable function comment. if the tag isn't in any of them,
+    # then it doesn't really matter since we're going to raise an exception anyways.
+    state_correct = internal.comment.decode(comment(ea, repeatable=repeatable))
+    state_wrong = internal.comment.decode(comment(ea, repeatable=not repeatable))
+    state_runtime = internal.comment.decode(function.comment(ea, repeatable=True))
```
arizvisa, line 1629: this call to `function.comment` needs to be guarded in case we're not in a function.

base/database.py
```diff
@@ -1530,29 +1540,57 @@ def tag(ea, key, value):
...
+    # in neither, then choose the comment type based on what we determined with
+    # the repeatable variable. we need to do special handling for runtime-linked
+    # functions because IDA uses repeatable function comments for some reason.
     ea = interface.address.inside(ea)
-    state = internal.comment.decode(comment(ea, repeatable=not repeatable))
-    state and comment(ea, '', repeatable=not repeatable) # clear the old one
-    state.update(internal.comment.decode(comment(ea, repeatable=repeatable)))
+    state_correct = internal.comment.decode(comment(ea, repeatable=repeatable))
+    state_wrong = internal.comment.decode(comment(ea, repeatable=not repeatable))
+    state_runtime = internal.comment.decode(function.comment(ea, repeatable=True))
```
arizvisa, line 1560: this call to `function.comment` needs to be guarded in case we're not within a function.

misc/hooks.py
```diff
@@ -213,9 +215,10 @@ def old_changed(cls, ea, repeatable_cmt):
         logging.debug(u"{:s}.old_changed({:#x}, {:d}) : Received comment.changed event for a {:s} comment at {:#x}.".format('.'.join((__name__, cls.__name__)), ea, repeatable_cmt, 'repeatable' if repeatable_cmt else 'non-repeatable', ea))
         cmt = utils.string.of(idaapi.get_cmt(ea, repeatable_cmt))
         fn = idaapi.get_func(ea)
+        rt, _ = internal.interface.addressOfRuntimeOrStatic(fn) if fn else (False, None)
 
         # if we're in a function, then clear our contents.
```
arizvisa, line 220: this comment needs to be updated.

arizvisa (commented):
almost there!

arizvisa (commented):
lgtm.

PR #77, opened 2020-11-21
Fixed the way internal submodules and api modules are imported by modifying their loader.

arizvisa:
This PR fixes the way that the submodule and api modules are imported so that their names and packages are correct. Previously, the submodules could actually be imported from anywhere. This was due to a bug that occurred due to the loaders manually tampering with the `sys.modules` dictionary. This actually allowed the `internal.exceptions` module to work recursively which shouldn't have been possible.

The was fixed by removing all logic that tampered with `sys.modules` instead only allowing the `imp.load_module` function to maintain the responsibility. After refactoring the logic to fix that issue, the `internal.exceptions` module needed to be fixed which involved fixing a type-o and replacing the recursive import with the `builtins` module. A whole bunch of documentation was also added to the loaders and the `idapythonrc.py` script itself.

### DIFF
--- a/base/_exceptions.py
+++ b/base/_exceptions.py
@@ -1,4 +1,4 @@
-import exceptions as E
+import builtins as E
 
 class UnicodeException(E.BaseException):
     """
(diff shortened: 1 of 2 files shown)


arizvisa (commented):
Looks good at the moment.

arizvisa (commented):
Final review prior to replacing `execfile` with `exec`.

PR #79, opened 2020-11-26
Fixed the hook for rebasing segments (`allsegs_moved`) so that it accesses the cache for both contents tags and externals properly.

arizvisa:
In order to update the tagcache, the `allsegs_moved` event is hooked with the `hooks.rebase` function. This function receives a vector containing the list of all the segments, and where to position them. What's supposed to happen is that we iterate through this vector to get each segment so that we can get the bounds of addresses that need to be translated. Once we have the bounds, we then iterate through our tag cache for each contents tag, and then for each globals tag. Once an item has been found in the cache, then we simply translate each address from its old value to the new value. This is done by `__rebase_function` for contents, and then `__rebase_globals` for global tags. The values yielded from this function are then used to update the user with what was performed.

Apparently, though, these functions had a couple of issues which resulted in them not working properly. One issue was that their implementation depended on the primary api which was doing bounds checks. As some values in the database have changed, these bounds checks would fail thus skipping over the relocation of contents tags. Another issue is that the way the contents tags were being identified was by taking the function's top address and using it to look up the cache to update. As the segment has already been moved, this would result in being unable to find the cache for a given function as it would be using the new function address instead of the old one. This issue actually failed silently even.

As externals are specially handled, a new helper method was added as `address.get_func_extern`. This method will return whether a given address has a function defined at it, and whether that address actually points to an import (or an external). To fix the other issue, both the implementations of `__rebase_function` and `__rebase_globals` will now yield the offset instead of the address that we relocated. This way the caller can receive the offset and then translate it into either the old or new address depending on what it needs.

Some off-by-ones were also identified which would result in contents tags being cached if they were on a segment boundary, but not within the segment. This was fixed as well.

### DIFF
--- a/misc/hooks.py
+++ b/misc/hooks.py
@@ -554,7 +554,7 @@ def __process_functions(percentage=0.10):
     It's intended to be called once the database is ready to be tampered with.
     """
     p = ui.Progress()
-    globals = set(internal.comment.globals.address())
+    globals = {item for item in internal.comment.globals.address()}
 
     total = 0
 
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
Didn't really review this, but I'm pretty sure it works because I wrote it some time ago. Python3 is kicking my ass, and I'm tired.

PR #81, opened 2020-11-26
Replaced all instances of the `logging.warn` function which is deprecated in Python3 with its new name `logging.warning`.

arizvisa:
In Python3, calling the `logging.warn` function to log a record raises a `DeprecationWarning`. In order to avoid this warning, we need to use its other alias, `logging.warning`.

This is the second part of the fix for issue #76, and is dependant on PR #80.

Issue #76, opened 2020-11-17
Python3 compatibility?


arizvisa:
At the moment, the plugin is only compatible with Python2 due to a variety of reasons. Fortuantely IDA supports both Py2 and Py3 plugins (albeit not at the same time). However, I imagine at some point IDA is going to switch to Py3 only which will break this plugin entirely.

Py3 3.10.x will introduce a pattern matching syntax into the language and so if we want to leverage this in order to wrap Hex-Rays, we'll need to support python3 at some point.

arizvisa:
Probably going to be releasing this PR along an alpha-beta release schedule as I'm not 100% sure how much stuff I might've broken, and it's going to take me a bit to fucking test everything.

arizvisa:
So, PRs #80, #81, #82, and #83 are all pre-requisites. This PRs do things like add wrappers, conditionals, etc. for some of the common-language stuff that was broken by Python3.

arizvisa:
Once all the pre-requisites get merged, then PR #84 should add support for Python3. This will require a ton of testing, and as such will be tested in stages. Fucking shoot me.

arizvisa:
PR #84 was finally merged. The next thing that will need to be done is to update the documentation.

arizvisa:
The documentation has been updated to include Python3 support, but it's still being built with Python2. This is pretty much the only hold-up. I'm working on issue #6, and probably after that I'll get back to figure out how to rewrite the parser so that it works on both Py2 and Py3...

arizvisa:
Currently doing a refactor as per #158...This will probably completely break the documenation generation process for some period of time, so I'm not sure when documentation will catch up. If only these Python autodoc projects would combine the documentation from multiple function definitions that have the exact same name....

arizvisa:
I'm pretty tired of writing documentation...

PR #82, opened 2020-11-26
Replaced all types that are either missing or deprecated in Python3 with suitable alternatives that are still compatible with Python2.

arizvisa:
Python3 removes types from the `types` module like `types.NoneType` and `types.TypeType`. This PR replaces `types.NoneType` with `None.__class__`, and `types.TypeType` with `builtins.type`. The `basestring` base class is also removed as there is now only one `str` class to rule them all. In order to retain Python2 compatibility, this class is replaced with `six.string_types`.

This PR is the third part of the fix for issue #76, and is dependant on PR #81.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -1341,7 +1341,7 @@ class ref_t(namedtypedtuple):
     optional as not all references will provide it.
     """
     _fields = ('address', 'opnum', 'reftype')
-    _types = (six.integer_types, (six.integer_types, types.NoneType), reftype_t)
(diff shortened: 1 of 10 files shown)


arizvisa (commented):
Looks good to me. Also, grepping says this is the truth. My eyes hurt.

arizvisa (commented):
This was also a simple rename for `basestring` and `types.NoneType` to `six.string_types`, and removal of the `long` type.. Looks like I was able to get everything.

arizvisa:
Switching the merge target to `master` since the PR it was based on was already merged.

PR #83, opened 2020-11-28
Replaced a number of `six` wrappers with builtins and comprehensions that are compatible with Python3.

arizvisa:
Python3 removes a number of dictionary iterator methods such as `dict.iterkeys`, `dict.itervalues`, and `dict.iteritems`. It has also deprecated a number of dictionary set methods such as `dict.viewkeys`, `dict.viewvalues`, and `dict.viewitems` since the `dict.keys`, `dict.values`, and `dict.items` methods now return the `dict_keys`, `dict_values`, and `dict_items` types (respectively). Although the `six` module provides wrappers for these methods that are compatible between both Python2 and Python3, we can actually accomplish compatibility without the `six` module wrappers.

This PR does exactly that by using the native `dict.keys`, `dict.values`, and `dict.items` methods when iterating through them. Whenever the situation arises where set operations need to be applied to the keys of a dictionary, then the keys are manually converted to a set prior to said operations. The `six.moves` module is deprecated in favor of the `builtins` module, and comprehensions are also used wherever possible (`set`, `list`, and `dict`) in order to avoid having to instantiate types using the builtin. Some Python3-incompatible operators are also fixed such as replacing the division operator (`/`) with the integer division operator (`//`).

This PR is the fourth part of the fix for issue #76, and is dependant on PR #82.

### DIFF
--- a/base/_utils.py
+++ b/base/_utils.py
@@ -533,7 +533,7 @@ class character(object):
     def to_hex(cls, integer):
         '''Given an integer, return the hex digit that it represents.'''
         if integer >= 0 and integer < 0x10:
-            return six.int2byte(integer + 0x30) if integer < 10 else six.int2byte(integer + 0x57)
+            return six.unichr(integer + 0x30) if integer < 10 else six.unichr(integer + 0x57)
         raise ValueError
 
     @classmethod
@@ -635,7 +635,7 @@ class character(object):
                     h, l = map(cls.of_hex, (hb.lower(), lb.lower()))
 
                     # coerce the digits into an ascii character and send the character to our result
-                    result.send(six.int2byte(
+                    result.send(six.unichr(
                         h * 0x10 |
                         l * 0x01 |
                     0))
--- a/base/_utils.py
+++ b/base/_utils.py
@@ -548,7 +548,7 @@ class character(object):
         # begin processing any input that is fed to us
         while True:
             ch = (yield)
-            n = six.byte2int(ch)
+            n = ord(ch)
 
             # check if character has an existing escape mapping
             if cls.mapQ(ch):
@@ -757,7 +757,7 @@ class string(object):
         """
         if isinstance(item, six.string_types):
             res = cls.escape(item, '\'')
-            if all(six.byte2int(ch) < 0x100 for ch in item):
+            if all(ord(ch) < 0x100 for ch in item):
                 return "'{:s}'".format(res)
             return u"u'{:s}'".format(res)
         elif isinstance(item, tuple):
--- a/base/_utils.py
+++ b/base/_utils.py
@@ -7,8 +7,7 @@ class for querying and filtering lists of things, support for aliasing
 functions, and a number of functional programming primitives (combinators).
 """
 
-import six
-from six.moves import builtins
+import six, builtins
 
 import os, logging, types, weakref
 import functools, operator, itertools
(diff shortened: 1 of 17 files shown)


base/_utils.py
```diff
@@ -728,7 +727,8 @@ def escape(cls, string, quote=''):
         transform = character.escape(res); next(transform)
 
         # iterate through each character, sending everything to res
-        for ch in iter(string or ''):
+        iterable = (ch for ch in (string or ''))
+        for ch in iterable:
```
arizvisa, line 731: This doesn't even need to be an iterator.

base/database.py
```diff
@@ -275,7 +274,8 @@ class functions(object):
 
     def __new__(cls):
         '''Returns a list of all of the functions in the current database.'''
-        return builtins.list(cls.__iterate__())
+        iterable = cls.__iterate__()
```
arizvisa, line 277: This doesn't need an intermediary assignment, and can just be a list-comprehension.

base/database.py
```diff
@@ -327,8 +327,8 @@ def list(cls, **type):
         listable = []
 
         # Some utility functions for grabbing frame information
-        flvars = lambda f: _structure.fragment(f.frame, 0, f.frsize) if f.frsize else iter([])
-        favars = lambda f: function.frame.args(f) if f.frsize else iter([])
+        flvars = lambda f: _structure.fragment(f.frame, 0, f.frsize) if f.frsize else (item for item in [])
+        favars = lambda f: function.frame.args(f) if f.frsize else (item for item in [])
```
arizvisa, line 331: These can all be lists, no need to be iterators.

base/database.py
```diff
@@ -4025,7 +4033,9 @@ def iterate(cls):
     @classmethod
     def length(cls):
         '''Return the number of marks in the database.'''
-        return len(builtins.list(cls.iterate()))
+        iterable = cls.iterate()
+        listable = [item for item in iterable]
```
arizvisa, line 4037: This doesn't need an intermediary `iterable`, and can just be a straight-up list-comprehension.

base/instruction.py
```diff
@@ -1587,15 +1586,15 @@ def memory(ea, op):
...
-        res = reversed(res) if database.config.byteorder() == 'little' else iter(res)
-        res = reduce(lambda agg, n: (agg*0x100)|n, six.iterbytes(res), 0)
-        sf = bool(res & maxval>>1)
+        res = get_bytes(addr, size) or b''
+        res = res[::-1] if database.config.byteorder() == 'little' else res[:]
+        res = functools.reduce(lambda agg, item: (agg * 0x100) | item, bytearray(res), 0)
+        sf = bool(res & maxval >> 1)
 
-        return armops.memory(long(addr), long(res-maxval) if sf else long(res))
+        return armops.memory(long(addr), long(res - maxval) if sf else long(res))
```
arizvisa, line 1597: Are these calls to `long` even necessary?

base/structure.py
```diff
@@ -905,7 +907,7 @@ def __getitem__(self, index):
         elif isinstance(index, six.string_types):
             res = self.by_name(index)
         elif isinstance(index, slice):
-            res = [self.__getitem__(i) for i in six.moves.range(self.parent.ptr.memqty)].__getitem__(index)
+            res = [self.__getitem__(i) for i in range(self.parent.ptr.memqty)].__getitem__(index)
```
arizvisa, line 910: No reason to explicitly call `object.__getitem__` here, when the native operator should just be used.

custom/tagfix.py
```diff
@@ -157,11 +157,12 @@ def contents(ea):
     # update addresses and tags to contents
     ui.navigation.set(ea)
     logging.debug(u"{:s}.contents({:#x}): Updating the name reference cache for the contents of function {:#x}.".format('.'.join([__name__]), ea, ea))
-    for k, v in six.iteritems(tags):
+
```
arizvisa, line 160: This is an extra newline that can likely be culled.

misc/tools.py
```diff
@@ -164,7 +162,7 @@ def checkmarks():
             pass
         continue
 
-    d = list(res)
+    d = res[:]
```
arizvisa, line 165: Check that `res` is a list that needs to be copied and make sure it's not an iterator.

misc/tools.py
```diff
@@ -181,8 +179,8 @@ def checkmarks():
         return
 
     for k, v in functions:
-        print >>sys.stdout, "{:#x} : in function {:s}".format(k, func.name(func.byAddress(k)))
-        print >>sys.stdout, '\n'.join( ("- {:#x} : {:s}".format(a, m) for a, m in sorted(v)) )
+        six.print_("{:#x} : in function {:s}".format(k, func.name(func.byAddress(k))), file=sys.stdout)
```
arizvisa, line 182: This call to `func.byAddress` should be using the non-aliased version, `func.by_address`, or `func.by`.

arizvisa (commented):
A bug and a couple of minor fixes..

base/instruction.py
arizvisa: Removed them.

arizvisa (commented):
This pass after finding non-iterated `map` and `zip` calls seems to be okay too.

arizvisa (commented):
Seems okay after swapping all division operator w/ integer division ones.

arizvisa (commented):
Okay, this refactored away all instances of `filter`, `itertools.imap`, and `itertools.ifilter`.

arizvisa (commented):
Okay. This has been painfully checked too.

arizvisa:
Modifying target branch to master since the PR it was based on has already been merged.

PR #84, opened 2020-12-01
Re-implemented a number of wrappers and refactored the tag-parsing trie in order to be compatible with Python3.

arizvisa:
Python3 does a number of things to break compatibility with Python2. Due to this, there's a number of capabilities in this plugin that are broken when trying to run them in Python3. Some of these things are that the the parameters used for constructing code, function, and method types are different. Modules now include a `ModuleSpec` under the `module.__loader__` property. Parameters can't be unpacked in function definitions, and tuples can't be compared anymore. Other than things like `int` and `long` being consolidated into a single type, and `bytes` and `str` being considered as exclusive types (deprecating `unicode`) there were a few changes to the opcodes used by Python bytecode.

This PR fixes these issues in two major parts. The first part is to implement a compatibility class in `internal.utils`. This compatibility class allows one to interact with both function and code types in a portable manner between regardless of whether Python2 or Python3 is being used. The other part was to refactor the trie used in `internal.comment` to use different encoders/decoders depending on whether Python2 or Python3 is being used. This might introduce some backwards-compatibility issues as unicode strings were being trusted to Python's `eval` and `repr` operators.

The other minor fixes that were done include things like manually unpacking tuple-parameters in the implementation rather than the definition, and implementing a wrapper around the tuple that's used by the multicase logic in the `internal.interface` module. This wrapper is necessary for the `heapq` that's used due to Python3 not being able to compare tuples anymore. Previously, the priority was passed as the first element in the `heapq` and thus would be used to identify which function case to dispatch to. The new implementation will now simply wrap the tuple, and if the comparison methods are called on the tuple then the implementation will compare the priority stored in the first element of both tuples.

Unfortunately some pythonic types ended up being removed as a result of Python3. Types composed of `long`, `unicode`, or `unichr` are now non-existent and thus the user will need to explicitly specify their element size as a tuple.

This PR is the final part (5th) of the fix for #76, and is dependant on PR #83.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -97,10 +97,12 @@ class typemap:
 
         stringmap = {
             chr:(idaapi.strlit_flag(), idaapi.STRTYPE_C),
-            unichr:(idaapi.strlit_flag(), idaapi.STRTYPE_C_16),
             str:(idaapi.strlit_flag(), idaapi.STRTYPE_C),
-            unicode:(idaapi.strlit_flag(), idaapi.STRTYPE_C_16),
         }
+        if hasattr(builtins, 'unichr'):
+            stringmap.setdefault(builtins.unichr, (idaapi.strlit_flag(), idaapi.STRTYPE_C_16))
+        if hasattr(builtins, 'unicode'):
+            stringmap.setdefault(builtins.unicode, (idaapi.strlit_flag(), idaapi.STRTYPE_C_16))
 
         ptrmap = { sz : (idaapi.off_flag() | flg, tid) for sz, (flg, tid) in integermap.items() }
         nonemap = { None :(idaapi.align_flag(), -1) }
@@ -122,20 +124,26 @@ class typemap:
 
         stringmap = {
             chr:(idaapi.asciflag(), idaapi.ASCSTR_TERMCHR),
-            unichr:(idaapi.asciflag(), idaapi.ASCSTR_UNICODE),
             str:(idaapi.asciflag(), idaapi.ASCSTR_TERMCHR),
-            unicode:(idaapi.asciflag(), idaapi.ASCSTR_UNICODE),
         }
+
+        if hasattr(builtins, 'unichr'):
+            stringmap.setdefault(builtins.unichr, (idaapi.asciflag(), idaapi.ASCSTR_UNICODE))
+        if hasattr(builtins, 'unicode'):
+            stringmap.setdefault(builtins.unicode, (idaapi.asciflag(), idaapi.ASCSTR_UNICODE))
 
         ptrmap = { sz : (idaapi.offflag() | flg, tid) for sz, (flg, tid) in integermap.items() }
         nonemap = { None :(idaapi.alignflag(), -1) }
 
     # lookup table for type
     typemap = {
-        int:integermap, long:integermap, float:decimalmap,
-        str:stringmap, unicode:stringmap, chr:stringmap, unichr:stringmap,
+        int:integermap, float:decimalmap,
+        str:stringmap, chr:stringmap,
         type:ptrmap, None:nonemap,
     }
+    if hasattr(builtins, 'long'): typemap.setdefault(builtins.long, integermap)
+    if hasattr(builtins, 'unicode'): typemap.setdefault(builtins.unicode, stringmap)
+    if hasattr(builtins, 'unichr'): typemap.setdefault(builtins.unichr, stringmap)
 
     # inverted lookup table
     inverted = {}
@@ -210,7 +218,7 @@ class typemap:
         if isinstance(pythonType, ().__class__):
             (t, sz), count = pythonType, 1
             table = cls.typemap[t]
-            flag, typeid = table[abs(sz) if t in {int, long, float, type} else t]
+            flag, typeid = table[abs(sz) if t in {int, getattr(builtins, 'long', int), float, type} else t]
 
         # an array, which requires us to recurse...
         elif isinstance(pythonType, [].__class__):
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -333,7 +333,7 @@ class prioritybase(object):
 
         # add the callable to our priority queue
         res = self.__cache__[target]
-        heapq.heappush(self.__cache__[target], (priority, callable))
+        heapq.heappush(self.__cache__[target], internal.utils.priority_tuple(priority, callable))
 
         # preserve a backtrace so we can track where our callable is at
         self.__traceback[(target, callable)] = traceback.extract_stack()[:-1]
@@ -363,7 +363,7 @@ class prioritybase(object):
         # If we aggregated some items, then replace our cache with everything
         # except for the item the user discarded.
         if state:
-            self.__cache__[target][:] = state
+            self.__cache__[target][:] = [internal.utils.priority_tuple(*item) for item in state]
 
         # Otherwise we found nothing and we can remove the entire target
         # from our cache.
@@ -531,7 +531,7 @@ class priorityhook(prioritybase):
 
         # unhook, assign our new method, and then re-hook
         with self.__context__():
-            method = types.MethodType(closure, self.object, self.__type__)
+            method = internal.utils.pycompat.method.new(closure, self.object, self.__type__)
             setattr(self.object, name, method)
         return True
 
@@ -543,7 +543,7 @@ class priorityhook(prioritybase):
         if not hasattr(self.object, name):
             cls, method = self.__class__, '.'.join([self.object.__class__.__name__, name])
             raise NameError("{:s}.disconnect({!r}, {!s}) : Unable to disconnect from the specified hook ({:s}).".format('.'.join([__name__, cls.__name__]), name, callable, method))
-        method = types.MethodType(closure, self.object, self.__type__)
+        method = internal.utils.pycompat.method.new(closure, self.object, self.__type__)
         setattr(self.object, name, method)
         return True
 
--- a/base/_utils.py
+++ b/base/_utils.py
@@ -665,8 +665,8 @@ class character(object):
         # whitespace characters as a set
         whitespace = { ch for ch in _string.whitespace }
 
-        # printable characters as a set
-        printable = { ch for ch in _string.printable } - whitespace
+        # printable characters as a set (spaces are the only whitespace that we consider as printable)
+        printable = { ch for ch in _string.printable } - whitespace | {u' '}
 
         # hexadecimal digits as a lookup
         hexadecimal = { ch : i for i, ch in enumerate(_string.hexdigits[:0x10]) }
@@ -743,7 +743,7 @@ class character(object):
                 result.send(ch)
 
             # check if character is printable (ascii)
-            elif isinstance(ch, str) and cls.asciiQ(ch):
+            elif isinstance(ch, six.string_types) and cls.asciiQ(ch):
                 result.send(ch)
 
             # check if character is a single-byte ascii
--- a/base/_utils.py
+++ b/base/_utils.py
@@ -902,7 +902,7 @@ class string(object):
     }
 
     @classmethod
-    def escape(cls, string, quote=''):
+    def escape(cls, string, quote=u''):
         """Escape the characters in `string` specified by `quote`.
 
         Handles both unicode and ascii. Defaults to escaping only
@@ -946,15 +946,19 @@ class string(object):
 
         # Python2 string types (str/bytes and unicode)
         if isinstance(item, six.string_types) and sys.version_info.major < 3:
-            res = cls.escape(item, '\'')
+            res = cls.escape(item.decode('latin1') if isinstance(item, bytes) else item, u'\'')
             if all(ord(ch) < 0x100 for ch in item):
-                return "'{:s}'".format(res)
+                return u"'{:s}'".format(res)
             return u"u'{:s}'".format(res)
 
-        # Python3 string types (bytes and str)
-        elif isinstance(item, (six.string_types, bytes)):
-            res = cls.escape(item, '\'')
-            return u"b'{:s}'".format(res) if isinstance(item, bytes) else u"'{:s}'".format(res)
+        # Python3 string types (str and bytes)
+        elif isinstance(item, six.string_types):
+            res = cls.escape(item, u'\'')
+            return u"'{:s}'".format(res)
+
+        elif isinstance(item, bytes):
+            res = cls.escape(item.decode('latin1'), u'\'')
+            return u"b'{:s}'".format(res)
 
         elif isinstance(item, tuple):
             res = map(cls.repr, item)
--- a/base/_utils.py
+++ b/base/_utils.py
@@ -820,7 +820,11 @@ class wrap(object):
     CO_FUTURE_BARRY_AS_BDFL     = 0x40000
     CO_FUTURE_GENERATOR_STOP    = 0x80000
 
-    import opcode, compiler.consts as consts
+    import opcode
+    if sys.version_info.major < 3:
+        import compiler.consts as consts
+    else:
+        import inspect as consts
 
     @classmethod
     def co_assemble(cls, operation, operand=None):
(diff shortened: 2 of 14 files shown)


base/_utils.py
```diff
@@ -122,7 +122,172 @@ def __cmp__(self, other):
...
+            return object.__self__.__class__
+
+        @classmethod
+        def function(cls, object):
+            return object.__func__
+
+    method = method_2x if sys.version_info.major < 3 else method_3x
+
 ### decorators
+class heap_package(object):
```
arizvisa, line 271: I think this should be renamed to something like `priority_tuple` or something...

base/_utils.py
```diff
@@ -122,7 +122,172 @@ def __cmp__(self, other):
...
+        def globals(cls, object):
+            return object.__globals__
+        @classmethod
+        def closure(cls, object):
+            return object.__closure__
+
+    function = function_2x if sys.version_info.major < 3 else function_3x
+
+    class code_2x(object):
+        def __new__(cls, object):
```
arizvisa, line 193: This should be under `pycompat.function.code`.

base/_utils.py
```diff
@@ -122,7 +122,172 @@ def __cmp__(self, other):
...
+        @classmethod
+        def unpack_extra(cls, object):
+            return ()
+        @classmethod
+        def new(cls, attributes, extra=()):
+            argcount, nlocals, stacksize, flags, code, consts, names, varnames, filename, name, firstlineno, lnotab, freevars, cellvars = attributes
+            return types.CodeType(*attributes)
+
+    class code_3x(code_2x):
+        def __new__(cls, object):
```
arizvisa, line 221: This should be under `pycompat.function.code`.

base/_utils.py
```diff
@@ -338,14 +504,16 @@ def F(*arguments, **keywords):
             #return f(*(arguments + tuple(w)), **keywords)
 
         # swap out the original code object with our wrapper's
-        f, c = F, F.func_code
+        f, c = F, pycompat.code(F)
```
arizvisa, line 507: If we abide by the comment, then this should be `pycompat.function.code(F)`.

base/_utils.py
```diff
@@ -122,7 +122,172 @@ def __cmp__(self, other):
     def __repr__(self):
         return "{:s}({:s})".format('Pattern', '|'.join(n.__name__ for n in self.type) if hasattr(self.type, '__iter__') else self.type.__name__)
 
+### compatibility namespace
+class pycompat(object):
+    class function_2x(object):
+        def __new__(cls, object):
```
arizvisa, line 128: This should be in its own namespace as `pycompat.method.function`.

base/_utils.py
```diff
@@ -122,7 +122,172 @@ def __cmp__(self, other):
...
+        @classmethod
+        def globals(cls, object):
+            return object.func_globals
+        @classmethod
+        def closure(cls, object):
+            return object.func_closure
+
+    class function_3x(function_2x):
+        def __new__(cls, object):
+            return object.__func__
```
arizvisa, line 162: This should be in its own namespace as `pycompat.method.function`.

base/_utils.py
```diff
@@ -193,10 +358,11 @@ def result(wrapped):
                 continue
 
             # everything is ok...so should be safe to add it
-            heapq.heappush(cache, (priority, (func, t_args, argtuple)))
+            heapq.heappush(cache, heap_package(priority, (func, t_args, argtuple)))
```
arizvisa, line 361: This `heap_package` name should be named `priority_tuple`.

base/_utils.py
```diff
@@ -269,7 +435,7 @@ def Fargsiter(names=args, values=parameters):
             return
 
         # Log any multicased functions that define type constraints for parameters which don't exist.
-        co = func.func_code
+        co = pycompat.code(func)
```
arizvisa, line 438: `pycompat.function.code(func)`.

base/_utils.py
```diff
@@ -359,56 +527,57 @@ def ex_function(cls, object):
         if isinstance(object, types.FunctionType):
             return object
         elif isinstance(object, types.MethodType):
-            return object.im_func
+            return pycompat.function(object)
```
arizvisa, line 530: `pycompat.method.function(object)`.

base/_utils.py
```diff
@@ -697,15 +866,25 @@ class string(object):
...
         return None if string is None else string.encode('utf8') if isinstance(string, unicode) else string
 
+    @classmethod
+    def passthrough(cls, string):
+        '''Handle strings to and from IDA in a tarnsparent fashion.'''
+        if not isinstance(string, (six.string_types, bytes, None.__class__)):
+            raise TypeError(string)
+        return None if string is None else string.decode('utf8') if isinstance(string, bytes) else string
+
+    of = of_2x if sys.version_info.major < 3 else passthrough
```
arizvisa, line 885: Document this please.

base/_utils.py
```diff
@@ -959,11 +1146,11 @@ def extract(cls, object):
 
         # if it's a method, then extract the function from its propery
         elif isinstance(object, types.MethodType):
-            return object.im_func
+            return pycompat.function(object)
```
arizvisa, line 1149: `pycompat.method.function(object)`.

base/_utils.py
```diff
@@ -976,12 +1163,12 @@ def extract(cls, object):
     @classmethod
     def arguments(cls, f):
         '''Extract the arguments from a function `f`.'''
-        c = f.func_code
-        count, iterable = c.co_argcount, (item for item in c.co_varnames)
+        c = pycompat.code(f)
```
arizvisa, line 1166: `pycompat.function.code(f)`.

base/_interface.py
```diff
@@ -325,7 +333,7 @@ def add(self, target, callable, priority):
 
         # add the callable to our priority queue
         res = self.__cache__[target]
-        heapq.heappush(self.__cache__[target], (priority, callable))
+        heapq.heappush(self.__cache__[target], internal.utils.heap_package(priority, callable))
```
arizvisa, line 336: This should be named `internal.utils.priority_tuple`.

base/_interface.py
```diff
@@ -355,7 +363,7 @@ def discard(self, target, callable):
         # If we aggregated some items, then replace our cache with everything
         # except for the item the user discarded.
         if state:
-            self.__cache__[target][:] = state
+            self.__cache__[target][:] = [internal.utils.heap_package(*item) for item in state]
```
arizvisa, line 366: This should also be named `internal.utils.priority_tuple`.

arizvisa (commented):
Please review the following name changes, and don't forget to do a proper check of the refactoring of `internal.comment` trie.

arizvisa (commented):
Okay...all intended changes look okay. Just need to manually verify tag encoding/decoding using the trie from the `internal.comment` module.

arizvisa:
While testing this PR, a huge number of issues were identified that aren't even related to Python3. Prior to fixing this PR, these will need to be fixed as well.

- [ ] internal/interface
  - [ ] +1483 implement a `switch_t.__str__` method as an alternative to `switch_t.__repr__`
  - [ ] +1483 register needs to be formatted not as {:s}, but as {!s}
  - [ ] +1763 this bounds_t should be imported into the root namespace as a core type
- [ ] base/function
  - [ ] +225 this `convention` function belongs under the `type` namespace.
  - [ ] +609 implement the `chunks.register` function for filtering a chunk.
  - [ ] +1321 this reference to the `B` variable is supposed to be `bb`.
  - [ ] +1596 implement a `frame.lvars.__new__` method to yield all frame members similar to the `frame.args.__new__` method
  - [ ] +1620 implement a `frame.regs.__new__` method to yield all register members similar to the `frame.args.__new__` method
- [ ] base/instruction
  - [ ] +170 ensure that the address is pointing to code and an actual instruction
  - [ ] +220 check that the operand number is within the count of operands
  - [ ] +307 this ops_state function should be returning a tuple of sets/states like op_state 
  - [ ] +343 add another case to ops_register that when given no parameters will return the indices for any individual registers
  - [ ] +379 could this op_repr implementation include an alias for op_str? will this conflict with another op_string implementation somewhere?
  - [ ] +972 this op_structure function should return the actual member or path instead of the boolean in its default condition.
  - [ ] +1083 check the operand number is within the count of operands for the instruction
  - [ ] +1172 this conditional check if the type is an enumeration seems wrong as the case is never hit...double check these conditionals to ensure they actually make sense and work properly.
- [ ] base/structure
  - [ ] +291 what does this `structure_t.down` function do again?
  - [ ] +478 verify that the name was actually assigned properly, and raise an exception if it failed. log a warning, or raise an exception...anything but silence.
  - [ ] +491 raise an exception, log something, if the comment isn't assignd properly
  - [ ] +491 add a case that removes the comment when its value is assigned as None
  - [ ] +608 implement a `structure_t.__str__()` method or something for Python3, but more importantly we should not be encoding its value to utf-8
  - [ ] +681 add a case that removes the comment when its value is assigned with None
  - [ ] +934 the `fnmatch.fnmatch` function just doesn't work anymore in Python3. 
  - [ ] +1349 we shouldn't be generally encoding comments as utf-8 when logging stuff.
  - [ ] +1387 for some reason when serializing/deserializing this instance, its member_t is missing its private .__index attribute. is there some way to phase these attributes out entirely?
  - [ ] +1493 implement a `member_t.bounds` method which will return the bounds of a field
  - [ ] +1522 if the member name doesn't get assigned properly (or exactly) raise an exception, log a warning, or something.
  - [ ] +1534 if the member comment doesn't get assigned or isn't exact, then log a warning, raise an exception, something.
  - [ ] +1534 if None is assigned as a comment, then remove it
  - [ ] +1590 sys.maxint doesn't exist anymore, so this comparison should go away in favor of something portable
  - [ ] +1683 there's no reason to encode this thing to utf-8
  - [ ] +1762 this seems to fail sometimes in python3 with an empy sequence
- [ ] base/enumeration
  - [ ] +180 this size() function should return the number of bytes
  - [ ] +180 implement a bits() function to return the number of bits
  - [ ] +201 this repr() function should display a comment if one exists
  - [ ] +442 maybe an alias can be implemented as owner()?
- [ ] base/database
  - [ ] +65 this config.info property should be initialized by a hook, or needs to be a function
  - [ ] +150 the config.type function is using str.translate which is deprecated in python3
  - [ ] +170 this config.processor function is previously declared and should be removed
  - [ ] +224 what is this config.main function for?
  - [ ] +256 this config.registers.segmentsize is returning bytes, but should return bits?
  - [ ] +603 the fnmatch.fnmatch function for the functions namespace is gone in python3
  - [ ] +648 the definition constraints for the read function should be a tuple, not just an interface.bounds_t
  - [ ] +755 the fnmatch.fnmatch function for the names namespace is gone in py3
  - [ ] +764 a non-parameter case should be implemented for the names.name function should be defined
  - [ ] +764 maybe an alias for names.symbol can be added
  - [ ] +814 in python3, the unicode type is gone so we can't test for it
  - [ ] +932 the implementation of search.iterate requires catching an exception so we can terminate the generator properly instead of dying w/ an exception. maybe log the exception that's raised.
  - [ ] +1117 do we really need this erase() function?
  - [ ] +1224 the fnmatch.fnmatch function for the entries namespace is gone in py3
  - [ ] +1635 why is the `__color__` tag missing from the tag() removal function
  - [ ] +1842 the fnmatch.fnmatch function for the imports namespace is gone in py3
  - [ ] +1849 for some reason the staticmethod `__format__` is not working in the regex matching definition
  - [ ] +1962 the module need to be pulled from the format name in `cls.__iterate__` since it's being extracted from the symbol name itself.
  - [ ] +2130 the definition constraints for the imports.iterate function should be a tuple, not just an interface.bounds_t
  - [ ] +2142 this blocks function should include a case for the interface.bounds_t (tuple, really).
  - [ ] +2523 if we're within a function, the regmatching for addess.prevregister should skip over alignment or non-code if within a function, or something?
  - [ ] +2585 if we're within a function, the regmatching for address.nextregister should skip over alignment or non-code if within a function, or something?
  - [ ] +2639 if delta == 0, then any modification or tampering with $sp should be identified by address.prevdelta
  - [ ] +2662 if delta == 0, then any modification or tampering with $sp should be identified by address.nextdelta
  - [ ] +2743 for some reason this address.prevbranch function doesn't seem to work in py3?
  - [ ] +2775 for some reason this address.prevbranch function doesn't seem to work in py3?
  - [ ] +3063 the usage of the `func` variable in `type.__new__` does not exist
  - [ ] +3178 the result of type.is_initialized is not inverted, so it only returns True when something is uninitialized...
  - [ ] +3363 is this type.is_label function a duplicate of type.has_label? what's the difference?
  - [ ] +3384 this line of code at the top of type.has_typeinfo seems to cancel the rest of the logic...why is this being kept?
  - [ ] +3423 the `type.is_structure` function should probably have an alias as `type.structureQ`
  - [ ] +3423 how does this `type.is_reference` function differ from the `type.has_reference` function up above?
  - [ ] +3594 the `type.structure.size` function should only return the structure size, not the item size. we should figure out the structure, and then return its size
  - [ ] +3625 this accessing the `database` module is bunk code and should be `xref` because we're in the database module...
  - [ ] +3640 this accessing the `database` module is bunk code and should be `xref` because we're in the database module...
  - [ ] +4056 remove this camelCased byIndex alias
  - [ ] +4068 remove this camelCased byAddress alias
  - [ ] +4313 rename this `.__del__` method with something like `.__delete__` because it might conflict
  - [ ] +4339 this code `or b''` should be a string as the api does not return bytes
  - [ ] +4357 rename this `.__del__` method with something like `.__delete__` because it might conflict
  - [ ] +4596 test out this `set.info` function on python3 because i don't know if it really works
  - [ ] +4618 this implementation of the set.unknown function needs to still return the size of the unknown bytes instead of 0 if its already undefined...
  - [ ] +4729 i don't think the `alignment` parameter here actually works and calculates the alignment properly
  - [ ] +4856 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +4874 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +4898 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +4924 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +4948 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +4976 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +5003 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +5032 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +5059 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +5088 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +5115 this size comparison should be checked with `type.unknownQ`, and then the size should be checked if it was changed.
  - [ ] +5154 don't match the exact size here, and check if the size fits. maybe log a warning event if things need to change
  - [ ] +5231 the `type` parameter constraint here should really be untyped because it's not a list, and can be a tuple.
  - [ ] +5515 if the size for `get.float` is invalid and doesn't fit, clamp it to the nearest size and log a warning event or something
  - [ ] +6045 the instruction operand calculation here is fucked as it can result in a negative value which means we're not translating it to the correct base address or whatever.

arizvisa:
Modified the merge target to "master" since all the other commits have been merged already.

arizvisa (commented):
I guess this has been reviewed.

base/database.py
```diff
@@ -3159,10 +3159,10 @@ class type(object):
...
     Some examples of using this namespace can be::
 
-        > print database.type.size(ea)
-        > print database.type.is_initialized(ea)
-        > print database.type.is_data(ea)
-        > length = database.t.array.length(ea)
+        > print( database.type.size(ea) )
+        > print( database.type.is_initialized(ea) )
+        > print( database.type.is_data(ea) )
+        > length = database.t.array.length(ea) )
```
arizvisa, line 3165: this is wrong.

arizvisa (commented):
ugh...

arizvisa (commented):
Okay. I think this fucker can be merged?

Issue #85, opened 2020-12-03
Deprecating the types used for the parameters in the `internal.netnode` module in favor of the `memoryview`.


arizvisa:
The `internal.netnode` module uses a very particular syntax in order to determine which netnode functions to use when reading or writing to the various netnode types within a database. This syntax is based on using `str`, `int`, and `buffer` in order to determine which function to use and what type it should return.

In Python3, these types have been deprecated and are thus completely missing from its builtins. This is due to Python3 replacing `str` with a new `bytes` type, and then removing `buffer` since it's really just a protocol that an instance is supposed to follow. This means that we'll need to introduce breaking changes to `internal.netnode` if we still want to allow a user to explicitly specify which type to get and set netnodes with.

This issue is just a placeholder for a PR.

arizvisa:
This is fixed by PR #86.

PR #86, opened 2020-12-03
Replace the types used in the `internal.netnode` module with types that are common between both Python2 and Python3.

arizvisa:
The `internal.netnode` module uses a very particular syntax in order to determine which netnode functions to use when reading or writing to the various netnode types within a database. This syntax is based on using `str`, `int`, and `buffer` in order to determine which function to use and what type it should return.

In Python3, these types have been deprecated and are thus completely missing from its builtins. This is due to Python3 replacing `str` with a new `bytes` type, and then removing `buffer` since it's really just a protocol that an instance is supposed to follow. This means that we'll need to introduce breaking changes to `internal.netnode` if we still want to allow a user to explicitly specify which type to get and set netnodes with.

Previously the `buffer` type was used for determining whether things like `idaapi.netnode.hashstr_buf` would be called, and the `str` type was used to determine whether `idaapi.netnode.valstr` would be called due to the name directly mapping to the api. In Python2, the `str` type is really an alias for `bytes` and so we were previously doing explicit tests for this using `six.string_types` in order to dispatch to the correct byte handler. As the semantics of this are different in Python3, we instead use the `bytes` type as an alias for `buffer`, and fallback to `str` so that we can mirror the semantics of Python2.

It was noticed that some of the netnode api would cull out all the trailing nulls (for strings), or would cull out only the last null (for a buffer). Due to this inconsistency, we now handle both `bytes` and `memoryview` by fetching the entire netnode's value, and then explicitly converting it to the type the user requested. This means that the correct way to guarantee that you set/get the actual bytes at a particular netnode is to now use a `memoryview` object. This type will internally fetch the entire value of a netnode, and then wrap it in a read-only `memoryview`. Then, to get its value it will need to be converted with either `memoryview.tobytes` or `memoryview.tolist`.

This PR fixes issue #85.

### DIFF
--- a/base/database.py
+++ b/base/database.py
@@ -4206,7 +4206,7 @@ class extra(object):
     @classmethod
     def __has_extra__(cls, ea, base):
         sup = internal.netnode.sup
-        return sup.get(ea, base) is not None
+        return sup.get(ea, base, type=memoryview) is not None
 
     @utils.multicase()
     @classmethod
@@ -4234,7 +4234,7 @@ class extra(object):
     def __count__(cls, ea, base):
         sup = internal.netnode.sup
         for i in six.moves.range(cls.MAX_ITEM_LINES):
-            row = sup.get(ea, base+i)
+            row = sup.get(ea, base + i, type=memoryview)
             if row is None: break
         return i or None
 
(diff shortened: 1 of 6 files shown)


base/database.py
```diff
@@ -4317,7 +4317,7 @@ def __get__(cls, ea, base):
             if count is None: return None
 
             # grab the extra commenta from the database
-            res = (idaapi.get_extra_cmt(ea, base+i) or '' for i in six.moves.range(count))
+            res = (idaapi.get_extra_cmt(ea, base+i) or b'' for i in six.moves.range(count))
```
arizvisa, line 4320: This needs to be spaced.

arizvisa (commented):
Ugh...I almost don't even fucking care anymore...

arizvisa (commented):
Okay. It fucking works.

Issue #41, opened 2019-11-27
Documentation: Include docs for switching IDA 7.4's IDAPython from Python3 to Python2


arizvisa:
IDA 7.4 includes the option to enable the Python3 or Python2 IDAPython plugin by default. It would be great to update the docs to reference how to enable Python2 for IDAPython in a user's installation.

A friend pasted me the following filenames that contain the plugin libraries to rename for his platform.
```
 1994  mv idapython.dylib idapython.3.disabled
 1996  mv idapython64.dylib idapython64.3.disabled
 1997  mv idapython.2.disabled idapython.dylib
 1998  mv idapython64.2.disabled idapython64.dylib
/Applications/IDA Pro 7.4/ida.app/Contents/MacOS/plugins
```

arizvisa:
This issue is fixed by PR #87. The changes will make their way into the online documentation during the next release (which will probably happen in an hour or so after I rebuild the docs).

arizvisa:
Closing this issue due to release being completed (new docs are at https://arizvisa.github.io/ida-minsc/).

PR #87, opened 2020-12-04
Added documentation for how to switch IDAPython from Python 3.x to Python 2.x after it's been installed.

arizvisa:
Python 2.x has been deprecated in favor of Python 3.x, and thus IDAPython has a Python 3.x version available for it. However due to Python 3.x introducing incompatible changes with 2.x, IDAPython allows you to choose which version of Python to use at the time of install. Unfortunately, there's not much (if any) documentation on what a user needs to do in order to downgrade their version of IDAPython to the 2.x version. Thus there doesn't appear to be a straightforward reference on what one needs to do in order to accomplish this after it's already been installed.

This PR adds a section and references to the documentation describing how one can downgrade their version of IDAPython to Python 2.x for all of the known platforms. A few other things like spacing and quotes were also fixed in the documents that were tampered with. 

This closes issue #41.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,10 +25,12 @@ and annotation issues can be performed with just a few lines of code. This
 should enable a user to write quick, hacky, temporary code that can be used
 to augment their reversing endeavors without distraction.
 
-This plugin is dependant upon IDAPython with Python 2.x, and supports all of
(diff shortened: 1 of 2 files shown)


arizvisa (commented):
I'm just skipping this review because writing documentation sucks. Anyways...merging this shit.

Issue #88, opened 2020-12-08
Some function cases take an interface.bounds_t as one of its parameters which isn't available without accessing internal modules.


arizvisa:
Some functions like `database.read`, `database.address.iterate`, and `functions.blocks` implement cases which can take an `interface.bounds_t` as their parameter. Unfortunately the `interface.bounds_t` is not available to users and thus this case is impossible to reach. There are also a number of other functions that can implement cases which take this parameter type.

This issue is a placeholder for a couple of the fixes that are listed at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754 which are prerequisites in order to merge PR #84.

PR #89, opened 2020-12-09
Added support for keyword parameters to the `interface.bounds_t` constructor, imported it into the root namespace, and improved a number of places where it's used.

arizvisa:
This PR modifes the case definitions in a number of functions which originally used `interface.bounds_t` to a tuple. The `interface.bounds_t` type was not available to users which made these cases unreachable unless a user accessed this internal type directly. These types in the case definition were changed to tuples which should mitigate users have to specify the `bounds_t` entirely. The constructor for the `interface.bounds_t` tuple was also improved to allow using keywords to construct it, and any users of the type were also improved. 

This closes issue #88, and is part of the prerequisites list at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754.

### DIFF
--- a/__root__.py
+++ b/__root__.py
@@ -99,7 +99,7 @@ Any = utils.PatternAny()
 del(utils)
 
 # some types that the user might want to compare with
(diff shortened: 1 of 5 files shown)


arizvisa (commented):
omfg I hate python...like soo much.

Issue #90, opened 2020-12-09
Potential register filtering cases, and numerous places that should format register_t in its short form.


arizvisa:
There's a couple of functions that can contain a register-filtering option which don't. This include things like the `function.chunks` namespace, etc. Also some places when displaying the `register_t` are not using the correct formatspec which results in the long-form being displayed instead of the short-form.

This is just a placeholder issue for a couple of fixes that are listed at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754 which are prerequisites in order to merge PR #84.

PR #91, opened 2020-12-09
Added cases to all the register filters which allow the user to not need to explicitly specify a register.

arizvisa:
This PR adds another case to each of the register filtering functions such as `instruction.ops_register`, or `function.register` in order to allow the user to not need to explicitly specify a register. A number of the functions in the `database.address` namespace were also modified to skip over non-code that was found within a function or code segment.

This closes issue #90, and is part of the prerequisites list at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754.

### DIFF
--- a/base/function.py
+++ b/base/function.py
@@ -1344,8 +1344,8 @@ class block(object):
     @classmethod
     def register(cls, bounds, reg, *regs, **modifiers):
         '''Yield each `(address, opnum, state)` within the block identified by `bounds` that uses `reg` or any one of the registers in `regs`.'''
-        B = cls.at(bounds)
-        return cls.register(B, reg, *regs, **modifiers)
+        bb = cls.at(bounds)
+        return cls.register(bb, reg, *regs, **modifiers)
     @utils.multicase(bb=idaapi.BasicBlock, reg=(basestring, interface.register_t))
     @classmethod
     def register(cls, bb, reg, *regs, **modifiers):
--- a/base/function.py
+++ b/base/function.py
@@ -523,6 +523,27 @@ class chunk(object):
             yield ea
         return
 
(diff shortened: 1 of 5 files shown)


base/function.py
```diff
@@ -1318,13 +1339,13 @@ def register(cls, reg, *regs, **modifiers):
...
         bb = cls.at(ea)
-        return cls.register(B, reg, *regs, **modifiers)
+        return cls.register(bb, reg, *regs, **modifiers)
     @utils.multicase(bounds=builtins.tuple, reg=(basestring, interface.register_t))
     @classmethod
     def register(cls, bounds, reg, *regs, **modifiers):
         '''Yield each `(address, opnum, state)` within the block identified by `bounds` that uses `reg` or any one of the registers in `regs`.'''
-        bb = cls.at(bounds)
-        return cls.register(bb, reg, *regs, **modifiers)
+        B = cls.at(bounds)
```
arizvisa, line 1347: uhh...what?

base/instruction.py
```diff
@@ -306,7 +295,7 @@ def ops_state():
 def ops_state(ea):
     '''Returns a tuple of for all the operands containing one of the states "r", "w", or "rw" describing how the operands are modified for the instruction at address `ea`.'''
     ea = interface.address.inside(ea)
-    f = feature(ea)
+    f = type.feature(ea)
     res = ( ((f&ops_state.read[i]), (f&ops_state.write[i])) for i in six.moves.range(ops_count(ea)) )
     return tuple((r and 'r' or '') + (w and 'w' or '') for r, w in res)
```
arizvisa, line 300: You forgot to fix this.

arizvisa (commented):
meh...

arizvisa (commented):
ugh...

Issue #92, opened 2020-12-09
The implementation of the `database.set.unknown` function does not always return the size of the unknown data.


arizvisa:
Various users of the `database.set.unknown` function depend on it returning the size of the unknown data when it's been set. However, when `database.set.unknown` is applied to data that is already unknown, this function will return 0 because no data has been marked as unknown. This breaks functions such as the integer types in `database.set.integer`, `database.set.float`, etc.

This issue is a placeholder for a couple of the fixes that are listed at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754 which are prerequisites in order to merge PR #84.

PR #93, opened 2020-12-09
Fixed the implementation of the `database.set.unknown` function to always return the size of its undefined data.

arizvisa:
This PR fixes a number of the functions within the `database.set.integer` and `database.set.float` namespaces that occurred due to them trusting that `database.set.unknown` would always return the size of its undefined data. The issue was that when `database.set.unknown` was applied to already undefined data, 0 would be returned due to the failure of `idaapi.del_items`. This is fixed by explicitly checking whether the data is unknown after being applied, and falling back to returning the item size if so.

This closes issue #92, and is part of the prerequisites list at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754.

### DIFF
--- a/base/database.py
+++ b/base/database.py
@@ -3161,7 +3161,7 @@ class type(object):
     @staticmethod
     def is_initialized(ea):
         '''Return true if the address specified by `ea` is initialized.'''
-        return type.flags(interface.address.within(ea), idaapi.FF_IVL) == idaapi.FF_IVL
+        return type.flags(interface.address.within(ea), idaapi.FF_IVL) != idaapi.FF_IVL
     initializedQ = utils.alias(is_initialized, 'type')
 
     @utils.multicase()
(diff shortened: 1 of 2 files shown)


base/database.py
```diff
@@ -5203,21 +5221,22 @@ def structure(cls, ea, type):
...
     def array(cls, type, length):
         '''Set the data at the current address to an array with the specified `length` and `type`.'''
         return cls.array(ui.current.address(), type, length)
-    @utils.multicase(ea=six.integer_types, type=builtins.list)
+    @utils.multicase(ea=six.integer_types)
     @classmethod
     def array(cls, ea, type):
         '''Set the data at the address `ea` to an array of the specified `type`.'''
-        return cls.array(ea, type, 1)
+        type, length = type if isinstance(type, builtins.list) else (type, 1)
```
arizvisa, line 5238: this needs to be a list...not a tuple.

base/database.py
```diff
@@ -4931,7 +4946,7 @@ def sint16_t(cls):
         def sint16_t(cls, ea):
             '''Set the data at address `ea` to a sint16_t.'''
             res = set.unknown(ea, 2)
-            if res != 2:
+            if not type.is_unkonwn(ea, 2) or res < 2:
```
arizvisa, line 4949: type-o.

base/database.py
arizvisa: ...or not. I take that back.

arizvisa (commented):
some more things for me to fix. :-/

arizvisa (commented):
I guess this is okay...

Issue #94, opened 2020-12-09
The `database.type.is_importref` and `database.type.is_globalref` functions are mistakenly self-referencing the database module.


arizvisa:
The implementation of `database.type.is_importref` function and its related functions are mistakenly referencing the `database` module. This is an error because it's residing within the `database` module which makes this set of functions completely unusable.

This issue is a placeholder for a couple of the fixes that are listed at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754 which are prerequisites in order to merge PR #84.

PR #95, opened 2020-12-09
Removed some duplicate code from the `database` module and fixed both the `names` and `imports` namespaces.

arizvisa:
This PR removes a number of definitions that were discovered to be duplicates, and removes some camelCased aliases as well as deprecated/unnecessary functions. The `names` and `imports` namespaces were also tweaked in order to better and more consistently extract the module name from their symbols. The `search` namespace was also significantly improved with a version that is compatible with Python3.

This closes issue #94, and is part of the prerequisites list at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754.

### DIFF
--- a/base/database.py
+++ b/base/database.py
@@ -805,13 +805,13 @@ class search(object):
         """
         radix = direction.get('radix', 0)
 
-        # convert the data directly into a string of base-10 integers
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
blah blah blah...

PR #97, opened 2020-12-09
Updated the structure module to raise exceptions and logs events when unable to assign names or properties.

arizvisa:
This PR updates the `structure` module so that when assigning a name or comment to one of the properties belonging to the `structure_t` or `member_t`, the assignment is actually validated. If the assignment completely fails, then an exception is raised. If the resulting assignment does not exactly correspond to what the user requested, then a log event is created. The enumeration module was also fixed so that any comments for an enumeration or its members are included in the result from the `enum.repr` function. 

This closes issue #88, and is part of the prerequisites list at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754.

### DIFF
--- a/base/enumeration.py
+++ b/base/enumeration.py
@@ -209,10 +209,10 @@ def mask(enum):
 def repr(enum):
     '''Return a printable summary of the enumeration `enum`.'''
     eid = by(enum)
-    w = size(eid)*2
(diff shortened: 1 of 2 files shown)


arizvisa (commented):
okay. merge this you fuck.

Issue #98, opened 2020-12-09
The `instruction` module doesn't seem to do a good job of checking its parameters specifically with regards to the operand number.


arizvisa:
As mentioned in the title, there's a number of functions within the `instruction` module that should be checking that the operand number is within the number of operands for a given instruction. This can be checked, and should be checked. The `instruction.op_structure` function is also returning a boolean for one of its cases when it should be returning the path to its member.

This issue is a placeholder for a couple of the fixes that are listed at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754 which are prerequisites in order to merge PR #84.

PR #99, opened 2020-12-10
Updated the `instruction` module to check the number of operands prior to determining their result.

arizvisa:
This PR adds a number of checks to the functions in the `instruction` module to ensure that a valid operand number was chosen. Checks were also added to the `mnemonic` function to ensure that it was only being used on valid code. The `instruction.op_refs` function was also fixed so that it would properly raise an exception when the chosen operand number is an enumeration, and the `instruction.op_structure` function will now always return a structure member path. 
 
This closes issue #98, and is part of the prerequisites list at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754.


arizvisa (commented):
Okay. This is the last motherfucking PR for now.

Issue #10, opened 2018-09-23
Feature: The "notepad" namespace in the ui module needs methods to modify its contents


arizvisa:
The `ui.notepad` namespace is very primitive and is thus pretty useless unless you want to hide or show the window. It'd be nice if some methods were exposed that would allow a user to programmatically modify its contents. Methods that modify its contents might require either doing some work with the Qt API, or poking around to see if the contents are in a netnode or something related.

arizvisa:
This is fixed by PR #100.

arizvisa:
Closing this issue as PR #100 was merged.

PR #100, opened 2020-12-12
Added some methods to the `ui.notepad` namespace which allows one to interact with the notepad window's contents.

arizvisa:
In a far away galaxy, the `notepad` namespace was added to the `ui` module. Unfortunately there were no useful methods that were included which would allow one to interact with the contents of the window which made this a pretty worthless namespace.

This PR re-implements the `notepad.open` function so that it will return the `QtWidgets.QPlainTextEdit` child by default which will allow someone to interact with it in its contents in its entirety. The `notepad.close` function was also implemented in order to contrast `notepad.open` by hiding it whenever the function is called.

A number of other functions were also implemented to interact with the context of the `QtWidgets.QPlainTextEdit` that is returned. This includes both getting and setting text with `notepad.get` and `notepad.set` (respectively), appending text (with `notepad.append`), and interacting with the `QtGui.QTextCursor` that is returned by using the `notepad.cursor` function.

This closes the long withstanding issue #10.

### DIFF
--- a/misc/ui.py
+++ b/misc/ui.py
@@ -497,6 +497,63 @@ class notepad(appwindow):
     __open__ = staticmethod(idaapi.open_notepad_window)
     __open_defaults__ = ()
 
+    @classmethod
+    def open(cls, *args):
+        '''Open or show the notepad window and return its ``QtWidgets.QPlainTextEdit`` widget.'''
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
sweet and simple.

arizvisa:
Merging this thingo.

Issue #45, opened 2019-12-02
No exception is raised when IDA fails to modify the property of a structure or its members


arizvisa:
When accessing a structure (`strucure_t`) or a frame (`func.frame`) there are a number of properties that are exposed which can be modified. Each member from the structure (`member_t`) can also contain more properties which can be modified. Each of these are decorated with the `@property` decorator.

If IDA fails to modify these properties as per the implementation, nothing is raised or returned indicating success/failure. Although the setters of these methods actually return the result, since these are treated as object properties their result cannot be captured in any way. Hence this can result in confusion when trying to modify a property and nothing appears to actually have happened.

This pattern exists in a number of places within the `structure` module:
- [x] `structure_t.name`
- [x] `structure_t.comment`
- [x] `structure_t.size`
- [x] `structure_t.index`
- [x] `member_t.name`
- [x] `member_t.comment`
- [x] `member_t.type`
 
It also seems like the following properties are mis-decorated. The `member_t.typeinfo` for example is mis-decorated with the `@type.getter` decorator.
- [x] `member_t.typeinfo`

arizvisa:
While fixing this issue, it might be a good time to re-consider the `structure.__instance__` cache and remove it. Again, why are we caching these structures exactly?

arizvisa:
> While fixing this issue, it might be a good time to re-consider the `structure.__instance__` cache and remove it. Again, why are we caching these structures exactly?

Actually I think I was caching these so that when you fetch the member from a frame, the member will inherit its offset from its frame. If that was the only reason I was caching this, then it'd be better to temporarily assign the offset to a new structure_t when accessing the member's type.

arizvisa:
We should also create (and raise) exceptions for when there isn't a database open. This will clean up some of the error messages that are raised when the user tries to run scripts on an incompletely loaded database.

arizvisa:
This issue might be fixed by PR #97 due to the list of issues that were mentioned in the preqrequisites list at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754. This needs to be double-checked during the testing required for PR #84.

arizvisa:
This issue should've been fixed by PR #101. The `structure.__instance__` cache has been left, and database checking has been abandoned because of the amount of work atm.

Issue #96, opened 2020-12-09
When failing to assign names or comments to the properties a structure or its members it may fail silently.


arizvisa:
The `structure` module implements wrappers around a structure and its members. This is so that one can interact with its parts by accessing its properties directly. When modifying the name or comment of one of these components, IDA's name validation may result in the name or comment not being assigned. This ends up failing silently without informing the user that their assignment may not have been applied.

This issue is a placeholder for a couple of the fixes that are listed at https://github.com/arizvisa/ida-minsc/pull/84#issuecomment-740769754 which are prerequisites in order to merge PR #84.

arizvisa:
This issue was fixed by PR #97. Closing it.

PR #101, opened 2020-12-12
Updated the `structure` module to raise exceptions and log events if an api call has failed to set a property.

arizvisa:
Previously when assigning a property or calling a setter function within the `structure` module, the result of the IDAPython API call was never checked. This resulted in assignments to a `structure_t` or `member_t` object failing silently. Some of these issues were actually mentioned in issue #96 (and solved by PR #97), but not all of them.

This PR goes through a number of the cases, refactors the way some of the API calls are made so that they're checked, and if they're being used to modify a property they are then compared against the value that was assigned. If the API call failed then an exception is raised, whereas if the API call did not assign exactly what the user requested then an logging event gets created.

This is related to PR #97, and closes issue #45.

### DIFF
--- a/base/structure.py
+++ b/base/structure.py
@@ -564,7 +564,7 @@ class structure_t(object):
         ok = idaapi.expand_struc(self.ptr, 0, size - res, True)
         if not ok:
             cls = self.__class__
-            logging.fatal(u"{:s}({:#x}).size({:+d}) : Unable to resize structure \"{:s}\" from {:#x} bytes to {:#x} bytes.".format('.'.join([__name__, cls.__name__]), self.id, size, utils.string.escape(self.name, '"'), res, size))
+            raise E.DisassemblerError(u"{:s}({:#x}).size({:+d}) : Unable to resize structure \"{:s}\" from {:#x} bytes to {:#x} bytes.".format('.'.join([__name__, cls.__name__]), self.id, size, utils.string.escape(self.name, '"'), res, size))
         return res
 
     @property
--- a/base/structure.py
+++ b/base/structure.py
@@ -579,7 +579,15 @@ class structure_t(object):
     @index.setter
     def index(self, idx):
         '''Set the index of the structure to `idx`.'''
-        return idaapi.set_struc_idx(self.ptr, idx)
+        res = idaapi.get_struc_idx(self.id)
+        if not idaapi.set_struc_idx(self.ptr, idx):
+            cls = self.__class__
+            raise E.DisassemblerError(u"{:s}({:#x}).index({:+d}) : Unable to modify the index of structure \"{:s}\" from {:d} to index {:d}.".format('.'.join([__name__, cls.__name__]), self.id, idx, utils.string.escape(self.name, '"'), res, idx))
+
+        res = idaapi.get_struc_idx(self.id)
+        if res != idx:
+            logging.info(u"{:s}({:#x}).index({:+d}) : The index ({:d}) that the structure was moved to does not match what was requested ({:d}).".format('.'.join([__name__, cls.__name__]), self.id, idx, idx, res))
+        return res
 
     @property
     def typeinfo(self):
--- a/base/structure.py
+++ b/base/structure.py
@@ -481,13 +481,13 @@ class structure_t(object):
         oldname = idaapi.get_struc_name(self.id)
         if not idaapi.set_struc_name(self.id, ida_string):
             cls = self.__class__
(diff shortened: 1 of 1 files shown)


base/structure.py
```diff
@@ -581,9 +577,17 @@ def index(self):
...
         '''Set the index of the structure to `idx`.'''
-        return idaapi.set_struc_idx(self.ptr, idx)
+        res = idaapi.get_struc_idx(self.id)
+        if not idaapi.set_struc_idx(self.ptr, index):
+            cls = self.__class__
+            raise E.DisassemblerError(u"{:s}({:#x}).index({:+d}) : Unable to modify the index of structure \"{:s}\" from {:d} to index {:d}.".format('.'.join([__name__, cls.__name__]), self.id, index, utils.string.escape(self.name, '"'), res, index))
+
+        res = idaapi.get_struc_idx(self.id)
+        if res != index:
+            logging.info(u"{:s}({:#x}).index({:+d}) : The index ({:d}) that the structure was moved to does not match what was requested ({:d}).".format('.'.join([__name__, cls.__name__]), self.id, index, index, res))
```
arizvisa, line 589: This logevent is backwards.

base/structure.py
```diff
@@ -561,10 +558,9 @@ def size(self):
     def size(self, size):
         '''Expand the structure to the new `size` that is specified.'''
         res = idaapi.get_struc_size(self.ptr)
-        ok = idaapi.expand_struc(self.ptr, 0, size - res, True)
-        if not ok:
+        if not idaapi.expand_struc(self.ptr, 0, size - res, True):
             cls = self.__class__
-            logging.fatal(u"{:s}({:#x}).size({:+d}) : Unable to resize structure \"{:s}\" from {:#x} bytes to {:#x} bytes.".format('.'.join([__name__, cls.__name__]), self.id, size, utils.string.escape(self.name, '"'), res, size))
+            raise E.DisassemblerError(u"{:s}({:#x}).size({:+d}) : Unable to resize structure \"{:s}\" from {:#x} bytes to {:#x} bytes.".format('.'.join([__name__, cls.__name__]), self.id, size, utils.string.escape(self.name, '"'), res, size))
```
arizvisa, line 563: This size needs to be verified.

arizvisa (commented):
Some things I missed and that need to be fixed.

base/structure.py
```diff
@@ -480,15 +479,16 @@ def name(self, string):
 
         # now we can set the name of the structure
         oldname = idaapi.get_struc_name(self.id)
-        ok = idaapi.set_struc_name(self.id, ida_string)
-        if not ok:
+        if not idaapi.set_struc_name(self.id, ida_string):
+            cls = self.__class__
             raise E.DisassemblerError(u"{:s}({:#x}).name : Unable to assign the specified name ({:s}) to the structure {:s}.".format('.'.join([__name__, cls.__name__]), self.id, utils.string.repr(ida_string), utils.string.repr(oldname)))
```
arizvisa, line 484: missed the parameter here.

base/structure.py
```diff
@@ -480,15 +479,16 @@ def name(self, string):
...
 
         # verify that the name was actually assigned properly
         assigned = idaapi.get_struc_name(self.id) or ''
         if utils.string.of(assigned) != utils.string.of(ida_string):
+            cls = self.__class__
             logging.info(u"{:s}({:#x}).name : The name ({:s}) that was assigned to the structure does not match what was requested ({:s}).".format('.'.join([__name__, cls.__name__]), self.id, utils.string.repr(utils.string.of(assigned)), utils.string.repr(ida_string)))
```
arizvisa, line 490: and here too.

base/structure.py
```diff
@@ -1654,15 +1668,19 @@ def tag(self, key, value):
         '''Set the tag identified by `key` to `value` for the member.'''
         state = self.tag()
         repeatable, res, state[key] = True, state.get(key, None), value
-        ok = idaapi.set_member_cmt(self.ptr, utils.string.to(internal.comment.encode(state)), repeatable)
+        if not idaapi.set_member_cmt(self.ptr, utils.string.to(internal.comment.encode(state)), repeatable):
+            cls = self.__class__
+            raise E.DisassemblerError(u"{:s}({:#x}).comment : Unable to assign the provided comment to the structure member {:s}.".format('.'.join([__name__, cls.__name__]), self.id, utils.string.repr(self.name)))
```
arizvisa, line 1673: missed another.

arizvisa (commented):
"Another round!" -jc

arizvisa (commented):
Looks good.

arizvisa:
merging this bish.

Issue #102, opened 2020-12-13
Documentation: the None.__class__ type is emitted to the documentation as-is instead of with its actual name.


arizvisa:
As a result of PR #82, the generated documentation now emits the `None` type as `None.__class__`. This is wrong and should be emitted as `NoneType`.

arizvisa:
The docs `function.blocks.subgraph` also is not handling its `exits` tuple correctly. Instead its rendered as "intlong or list or tuple or set".

arizvisa:
This was fixed by PR #103...which was just merged. Closing...

PR #103, opened 2020-12-13
Added support for nested tuples, stringifying `None.__class__`, and some of the newer types.

arizvisa:
Due to PR #82, a number of aspects of the document parser ended up generating incorrect markdown. Some examples are that the `types.NoneType` being renamed to `None.__class__` emitted "None.\_\_class\_\_" during stringifcation instead of "None". Nested tuples also resulted in being joined together without the " or " separator. This also affected usage of some of the types from the `builtins` module which apparently use the type `future.types.newlist.newlist` to represent the `builtins.list` type.

To fix these numerous issues, the document parser (docs/docparse.py) was modified to stringify `None.__class__` to "None" and add a case for `future.types.newlist.newlist` to emit "list". All instances of `basestring` were promoted to `six.string_types` where possible, and all nested tuples were flattened into non-nested tuples.

This fixes issue #102.

### DIFF
--- a/docs/docparse.py
+++ b/docs/docparse.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import six
+import six, builtins
 import functools, operator, itertools, types
 import argparse, logging
 import ast, contextlib
@@ -42,8 +42,7 @@ class evaluate(object):
         return res
 
     ## Miscellaneous internal types
-    import six, types
-    from six.moves import builtins
+    import six, types, builtins
 
     class structure:
         structure_t = 'structure_t'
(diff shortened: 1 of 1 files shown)


docs/docparse.py
```diff
@@ -62,23 +61,23 @@ class stringify(object):
...
         chr: 'chr',
         callable: 'callable',
         tuple: 'tuple',
         set: 'set',
-        types.NoneType: 'None',
+        None.__class__: 'None',
     }
     stringmap = {
         'types.NoneType' : 'None',
+        'None.__class__' : 'None',
```
arizvisa, line 75: I love how I have no idea what I'm doing here...

arizvisa (commented):
HEh.

Issue #104, opened 2021-01-06
Unable to navigate the netnode types in reverse using the `utils` functions from the `internal.netnode` module.


arizvisa:
The `internal.netnode` module provides the `utils` namespace for simplifying the iteration through the different types of netnodes exposed by IDA. This is an internal namespace and uses `utils.falt`/`utils.ralt`, `utils.fsup`/`utils.rsup`, `utils.fhash`/`utils.rhash`, and `utils.fchar`/`utils.rchar` to iterate through the respective types.

The reverse-iteration variation of the functions are actually defined incorrectly and thus do not actually work. This results in an exception, `TypeError: netnode_hashprev expected at least 2 arguments, got 1` being raised upon usage.

arizvisa:
This is fixed by PR #105.

arizvisa:
Closing this issue as PR #105 was merged.

PR #105, opened 2021-01-06
Fixed an issue with reverse iteration of the netnode types in the `internal.netnode` module, and included a significant amount of documentation.

arizvisa:
The `internal.netnode` module is responsible for providing a number of utilities that are used to interact with a netnode within an IDA database. Most of the functionality within this module is designed to abstract iterating through the available types for a netnode. IDA actually provides ways for navigating forwards and backwards through these types which are also included by the module.

The `utils` namespace within the module provides wrappers for using IDAPython's API to iterate through all of the different types in both directions. The reverse direction, however, was being passed the wrong parameters which resulted in the functions not actually working. This PR fixes those incorrect parameters, and fully documents the module so that displaying the documentation with the `help` keyword actually works.

This closes issue #104.

### DIFF
--- a/base/_netnode.py
+++ b/base/_netnode.py
@@ -195,7 +195,7 @@ class utils(object):
         return
     @classmethod
     def ralt(cls, node):
-        for item in cls.valriter(node, netnode.altfirst, netnode.altprev, netnode.altnext, netnode.altval):
+        for item in cls.valriter(node, netnode.altfirst, netnode.altlast, netnode.altprev, netnode.altval):
             yield item
         return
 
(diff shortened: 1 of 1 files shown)


base/_netnode.py
```diff
@@ -303,158 +339,209 @@ def set(cls, nodeidx, value):
...
         return '\n'.join(res)
 
-### node sup iteration
+### node supval : sparse array[integer] = str * 1024
 class sup(object):
-    '''Sparse array[int] of 1024b strings'''
+    """
+    This namespace is used for interacting with the sparse array stored
+    within a given netnode. This sparse array is used to store bytes,
+    and is referred to by IDA as an "supval".
```
arizvisa, line 475: this "an" should be "a".

arizvisa (commented):
minor type-o to fix.

arizvisa (commented):
looks okay and ready to merge.

Issue #107, opened 2021-01-07
If a hashval netnode contains an empty string as the key, the api is unable to iterate through the keys and returns an empty list.


arizvisa:
When iterating through a netnode's hashval using the `netnode.fhash` function from the `internal.netnode` module, in some cases it will return an empty list. This happens despite there definitely being more than one hash key defined, and the key that triggers this is the empty string ('').

It's suspected that this is because IDAPython is incorrectly returning the empty string as `None`. Because `None` is being returned, the loop which iterates through all of the hashkeys will prematurely terminate as IDAPython uses `None` to return a failure. This implies that IDAPython is using truth to compare instead of `is` when comparing to `None`. To workaround this mistake, we might be able to check with `hashval` to determine if there's a value stored for the empty string. We can also check both `hashfirst` and `hashlast` to determine if there are no hashvals defined so that only then we can terminate the generator.

On another note, the `utils.rhash` function doesn't work properly due to it always returning the first key that was found instead of starting with the last one as it should.

arizvisa:
This issue is being handled by PR #108.

arizvisa:
PR #108 was merged which fixes this issue.

PR #108, opened 2021-01-07
Added special handling to the internal.netnode utilities when iterating through hash keys so that the empty key is supported properly despite its mishandling by IDAPython.

arizvisa:
IDAPython's netnode API provides a number of functions for enumerating the different types (supval, altval, and hashval) for a given netnode. These functions essentially provide the functionality to get the first, last, next, and previous values for the respective netnode type. When an error occurs within any of these functions, the `None` value is typically returned. Inside the `internal.netnode` module are a number of wrappers that are designed to simplify this interface, and make it consistent for all of the available types.

When iterating through the hashval type for a netnode, one would typically use `netnode.hashfirst` and `netnode.hashlast` to determine the start and end for all of the hashvals. Then to determine the items in between, the `netnode.hashnext` or `netnode.hashprev` would be used depending on whether the caller has chosen to start at the last hashval key, or the first hashval key. Thus the implementations of `utils.hfiter` and `utils.hriter` use each of these functions in a generic manner to yield each key along with the value of the key for every single available hashval in the netnode.

There's an issue, however, with IDAPython's implementation of `netnode.hashfirst` and `netnode.hashlast`. It seems that when an empty key is intended to be returned from either of these functions, the function will instead return `None`. As `None` is used to represent an error (or information that is missing), this results in the `utils.hfiter` and `utils.hriter` functions conidering that the netnode has no hashvals associated with it. I'm suspecting that this is because IDAPython (or whatever) is verifying that the key exists by checking for the hashval key's truthiness. As an empty string ('') is `False`, these APIs will instead return `None`.

To remedy this mishandling, rather than using `None` to check for an error, we prefix each iteration by checking to see if the start key (from `netnode.hashfirst`) is equivalent to the last key (from `netnode.hashlast`). If this is True and either of the keys are `None`, then the value for the empty string ('') is checked to see if it was defined. If if wasn't, then there really is no hashvals defined for the netnode, and we can terminate the iteration. Otherwise, if the empty string ('') has a value associated with it, then we ensure that we replace all instances of `None` with the empty string so that all the other logic should be the same.

This fixes issue #107.

### DIFF
--- a/base/_netnode.py
+++ b/base/_netnode.py
@@ -617,7 +617,7 @@ class hash(object):
             l1, l2 = 0, 2
 
         for index, key in enumerate(cls.fiter(nodeidx)):
-            value = "{:<{:d}s} : default={!r}, memoryview={!r}, bytes={!r}, int={:#x}({:d})".format("{!r}".format(cls.get(nodeidx, key)), l2, cls.get(nodeidx, key, None), cls.get(nodeidx, key, memoryview), cls.get(nodeidx, key, bytes), cls.get(nodeidx, key, int), cls.get(nodeidx, key, int))
+            value = "{:<{:d}s} : default={!r}, bytes={!r}, int={:#x}({:d})".format("{!r}".format(cls.get(nodeidx, key)), l2, cls.get(nodeidx, key, None), cls.get(nodeidx, key, bytes), cls.get(nodeidx, key, int), cls.get(nodeidx, key, int))
             res.append("[{:d}] {:<{:d}s} -> {:s}".format(index, key, l1, value))
         if not res:
             raise internal.exceptions.MissingTypeOrAttribute(u"{:s}.repr({:#x}) : The specified node ({:x}) does not have any hashvals.".format('.'.join([__name__, cls.__name__]), nodeidx, nodeidx))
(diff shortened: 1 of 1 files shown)


base/_netnode.py
```diff
@@ -184,22 +184,34 @@ def valriter(cls, node, first, last, prev, val):
     def hfiter(cls, node, first, last, next, val):
         '''Iterate through all of the hash values for a netnode in order, and yield the (item, value) for each item that was found.'''
         start, end = first(node), last(node)
-        if val(node, start) is None: return
-        yield start, val(node, start)
+
+        # if start is not defined, its the same as end, and there's no value
```
arizvisa, line 188: forgot an apostrophe here. whee...grammar.

base/_netnode.py
```diff
@@ -184,22 +184,34 @@ def valriter(cls, node, first, last, prev, val):
...
         return
 
     @classmethod
     def hriter(cls, node, first, last, prev, val):
         '''Iterate through all of the hash values for a netnode in reverse order, and yield the (item, value) for each item that was found.'''
         start, end = first(node), last(node)
-        if val(node, start) is None: return
-        yield end, val(node, end)
+
+        # if end is not defined, its the same as start, and there's no value
```
arizvisa, line 205: ditto. missing apostrophe.

arizvisa (commented):
minor grammar issues.

arizvisa (commented):
lgtm. ready to merge this bish.

Issue #109, opened 2021-01-08
Unable to deserialize structure members due to an AttributeError being raised


arizvisa:
When deserializing a `structure_t` that was serializing via pickle or anything else, an exception gets raised by the `member_t.__setstate__` method. This looks like the following at the REPL.

```
Python>b=pickle.loads(a)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib64/python2.7/pickle.py", line 1388, in loads
    return Unpickler(file).load()
  File "/usr/lib64/python2.7/pickle.py", line 864, in load
    dispatch[key](self)
  File "/usr/lib64/python2.7/pickle.py", line 1223, in load_build
    setstate(state)
  File "/home/user/work/ida-minsc/base/structure.py", line 1491, in __setstate__
    logging.info(u"{:s}({:#x}): Field at {:+#x} of structure ({:s}) contains the same name \"{:s}\".".format('.'.join([__name__, cls.__name__]), self.id, ofs, parentname, utils.string.escape(name, '"')))
  File "/home/user/work/ida-minsc/base/structure.py", line 1521, in id
    return self.ptr.id
  File "/home/user/work/ida-minsc/base/structure.py", line 1517, in ptr
    return self.__parent.ptr.get_member(self.__index)
AttributeError: _member_t__index
```

This is actually happening because of a recursion issue when trying to display an error message.

arizvisa:
This is fixed by PR #110.

arizvisa:
Closing this issue as PR #110 was merged.

PR #110, opened 2021-01-08
Fixed the structure.member_t class so that the properties are deserialized in the correct order and the type information is preserved.

arizvisa:
The `structure_t` and `member_t` classes are used to wrap structures that are defined within an IDA database. Both of these classes can actually be serialized because they implement the pickle protocol. When deserializing them, however, an error can occur which results in them recursively accessing themselves in order to format the error message. This results in an exception due to the classes not being properly instantiated at the time of formatting.

This PR fixes this issue in two ways. The first way is that all of the private properties that are defined for each class are now protected so that they're easier to access. The other way is that the log events that are created when dealing with naming issues were modified to not attempt to recurse into the object that is being instantiated. The time the properties are created are also guaranteed to be performed after the member has been added to the parent structure. The identifier is also checked in the `member_t` when the parent structure is created in order to raise an exception when the method is unable to create the parent structure.

The protocol for serializing the `member_t` object will also now include the actual type information (`idaapi.tinfo_t`) so that after the type has been constructed, the type information will be applied to the member. This was done by changing the original `idaapi.opinfo_t` into a tuple, and then using the `idaapi.tinfo_t.serialize` method to pack the type information. If the type wasn't applied properly, a warning is logged and deserialization will continue as if there was no type information packed.

This closes issue #109.

### DIFF
--- a/base/structure.py
+++ b/base/structure.py
@@ -907,19 +907,20 @@ class members_t(object):
         > result = st.members.by(offset=0x2a)
 
     """
-    __slots__ = ('__parent', 'baseoffset')
+    __slots__ = ('__parent__', 'baseoffset')
 
     # members state
     @property
     def parent(self):
         '''Return the ``structure_t`` that owns this ``members_t``.'''
-        return self.__parent
+        return self.__parent__
     @property
     def ptr(self):
         '''Return the pointer to the ``idaapi.member_t`` that contains all the members.'''
-        return self.__parent.ptr.members
+        parent = self.parent
+        return parent.ptr.members
     def __init__(self, parent, baseoffset=0):
-        self.__parent = parent
+        self.__parent__ = parent
         self.baseoffset = baseoffset
 
     def __getstate__(self):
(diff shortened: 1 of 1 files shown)


base/structure.py
```diff
@@ -1435,86 +1456,113 @@ class member_t(object):
...
+        cls, fullname = self.__class__, '.'.join([parentname, name])
 
         # get the structure owning the member by the name we stored
+        # creating it if necessary.
         res = utils.string.to(parentname)
         identifier = idaapi.get_struc_id(res)
         if identifier == idaapi.BADADDR:
-            cls = self.__class__
-            logging.info(u"{:s}({:#x}) : Creating member for structure ({:s}) at offset {:+#x} named \"{:s}\" with the comment {!r}.".format('.'.join([__name__, cls.__name__]), identifier, parentname, ofs, utils.string.escape(name, '"'), cmtt or cmtf or ''))
+            logging.info(u"{:s}({:#x}) : Creating member for structure ({:s}) at offset {:+#x} named \"{:s}\" with the comment {!r}.".format('.'.join([__name__, cls.__name__]), identifier, res, ofs, utils.string.escape(name, '"'), cmtt or cmtf or ''))
```
arizvisa, line 1489: this should not be accessing the `res` variable.

base/structure.py
```diff
@@ -1435,86 +1456,113 @@ class member_t(object):
...
             identifier = idaapi.add_struc(idaapi.BADADDR, res)
-        self.__parent = parent = __instance__(identifier, offset=0)
 
-        # extract the attributes of the member
+        if identifier == idaapi.BADADDR:
+            raise E.DisassemblerError(u"{:s}({:#x}) : Unable to get structure ({:s}) for member at offset {:+#x} named \"{:s}\" with the comment {!r}.".format('.'.join([__name__, cls.__name__]), identifier, res, ofs, utils.string.escape(name, '"'), cmtt or cmtf or ''))
```
arizvisa, line 1493: this also should not be accessing `res` since it's a string.

arizvisa (commented):
mistakes were made.

arizvisa (commented):
It's late. I'm tired. So, I'm merging this.

Issue #106, opened 2021-01-07
Issues on IDA 6.8 when trying to execute `database.get.info` due to the `database.config.__nw_init_info_structure__` notification not being executed.


gool123456:
Hi, @arizvisa , I tested another problem at old ida,  the command will have an error in the string or array data :

at byte_3138BC    :

.text:003138BC byte_3138BC     DCB 0
.text:003138BD                 DCB 0x20
.text:003138BE                 DCB    0
.text:003138BF                 DCB 0x6D ; m
.text:003138C0                 DCB    0
.text:003138C1                 DCB    4
.text:003138C2                 DCB    0
.text:003138C3                 DCB    0
.text:003138C4                 DCB 0x11
.text:003138C5                 DCB    0
.text:003138C6                 DCB 0x6D ; m
.text:003138C7                 DCB    0

I shell command **db.get.info()** , this show error , I want see this ea type,  
So I want to define this as an array and **display Indexs**. Is there such a method in ida-minsc? because you can find this option manually in ida with Ctrl+*.

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 290, in database.info
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 5027, in info
    return cls.info(ui.current.address())
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 290, in database.info
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 5032, in info
    return type(ea)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 290, in database.__new__
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 2853, in __new__
    res = idaapi.guess_tinfo(ea, ti) if idaapi.__version__ < 7.0 else idaapi.guess_tinfo(ti, ea)
  File "D:\tools\IDA_Pro_v6.8_and_Hex-Rays_Decompiler_(ARM,x64,x86)_Hskyre\python\idaapi.py", line 35611, in guess_tinfo
[...]

PR #111, opened 2021-01-08
Fixed some issues that would result in the incorrect API being used on older versions of IDA and would prevent some of the necessary hooks from working properly.

arizvisa:
In order for the plugin to provide a number of its capabilities to IDA, the plugin hooks many parts of IDA in order to track or transform itself according to what the user's needs might be. On newer versions of IDA, this is done with IDAPython's hooking capabilities. On older versions of IDA, however, this doesn't exist and so the plugin accomplishes its needs by using notification hooks. These notifications are not as granular as the regular hooks, but should get most things resolved.

When opening up a database, the plugin will hook processor selection via the `IDP_Hooks.ev_newprc` function. This way the plugin can swap its internal architecture state (representing registers and operand decoding) for the correct one depending on what processor IDA has selected. Since older versions of IDA don't provide this hook, we use a notification hook on the `NW_OPENIDB` notification with below-average priority which should allow the plugin to update its namespace so that its dependencies will also work properly.

As the hooks themselves are very similar, most of the wrapper hooks and notifications will simply call the actual hook after determining the correct parameters so that they can still work as they were originally intended. The `config.__nw_init_info_structure__` hook will actually call into the original `config.__init_info_structure__` after it has determined its parameters. Unfortunately, the implementation of that hook has misspelled the function that it calls into as `config.__init_into_structure__` and thus an exception will get raised and the plugin will fail to initialize its architecture for the loaded database.

This PR corrects that misspelling, and fixes a number of compatibility issues with the plugin on older versions of IDA. Another example of some of the other compatibility issues that needed to be fixed are the usage of the `idaapi.guess_tinfo` function which should be `idaapi.guess_tinfo2` on older versions of IDA as demonstrated in commit 7af2805b9d9d3a38d0b6065af4419e60917ab3d7. This same issue was discovered in the `function` module and was also fixed by this PR.

This intends to solve issue #106.

### DIFF
--- a/base/database.py
+++ b/base/database.py
@@ -89,7 +89,7 @@ class config(object):
     def __nw_init_info_structure__(cls, nw_code, is_old_database):
         logging.debug(u"{:s}.__nw_init_info_structure__({!s}) : Received notification to initialize information structure for database.".format('.'.join([__name__, cls.__name__]), ', '.join(map("{!r}".format, [nw_code, is_old_database]))))
         idp_modname = idaapi.get_idp_name()
-        return cls.__init_into_structure__(idp_modname)
+        return cls.__init_info_structure__(idp_modname)
 
     @classmethod
     def filename(cls):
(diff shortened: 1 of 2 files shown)


arizvisa (commented):
Appears sane. Merging it.

Issue #112, opened 2021-01-14
Case-sensitive vs. Case-insensitive matching dependant on the platform being used?


arizvisa:
It seems that the matching classes used by things like `database.functions`, `database.imports`, `database.entries`, etc. only does exact matching when running on the non-windows platform. These semantics do not correspond to what the documentation claims.

Turns out that the case-insensitive matching only works on the Windows platform due to the usage of the `fnmatch.fnmatch` function which does case-sensitive matching on posixy platforms, and case-insensitive matching on windows.

arizvisa:
This is fixed by PR #113. Linking it.

PR #113, opened 2021-01-15
Implemented another matcher type, and refactored the combinators used by it to optimize them for performance.

arizvisa:
When using the matcher utility to match various parts of a database with a glob, the matcher's case-sensitivity will vary depending on whether the platform being used is either posix or windows. This is because the combinator that is used to construct the matching predicate is based on the `fnmatch` python module. According to the documentation for the module, the implementation of the `fnmatch` python module is intended to be used only for matching paths on a filesystem. Thus when on a platform where the filesystem is case-sensitive, this symptom will manifest itself when performing a query on the database. On windows, however, these matchers seem to work as originally intended.

This PR fixes the main issue by using the `fnmatch.translate` function to transform the glob into a regular expression, and then the regular expression is applied to the queried items with the `IGNORECASE` flag set. For the string-comparison matchers, the `operator.methodcaller('lower')` function is inserted into the combinator in order to normalize the strings prior to performing the actual comparison. During this modification, it was noticed that a couple of the combinators could be significantly improved in order to significantly improve the performance when matching against a decent amount of data.

One common optimization is to actually take advantage of compiled regular expressions, as the generated regular expression is being parsed prior to each comparison which is entirely unnecessary as the regular expression does not change between comparisons. Thus to take advantage of this capability, the internal `utils.matcher` class was extended to allow the caller to provide their own combinator without any unnecessary transformations. After this addition, all the relevant matchers were updated to use a more minimalistic combinator which would execute any required calculations or compilations prior to performing the requested match on the chosen data.

This fixes issue #112.


base/_utils.py
```diff
@@ -432,21 +432,24 @@ class matcher(object):
...
-        return lambda o: tuple(x(o) for x in res) if len(res) > 1 else res[0](o)
+    def __attrib__(self, *attributes):
+        if not attributes:
+            return lambda item: item
+        res = [(operator.attrgetter(callable_or_attribute) if isinstance(callable_or_attribute, six.string_types) else callable_or_attribute) for callable_or_attribute in attributes]
+        return fcompose(*res) if len(res) > 0 else res[0]
     def attribute(self, type, *attribute):
         attr = self.__attrib__(*attribute)
-        self.__predicate__[type] = lambda v: fcompose(attr, functools.partial(functools.partial(operator.eq, v)))
+        self.__predicate__[type] = lambda target: fcompose(attr, functools.partial(functools.partial(operator.eq, target)))
```
arizvisa, line 442: This should probably be combinator based...but I really don't care to fuck with it right now.

base/_utils.py
```diff
@@ -432,21 +432,24 @@ class matcher(object):
...
     def mapping(self, type, function, *attribute):
         attr = self.__attrib__(*attribute)
         mapper = fcompose(attr, function)
-        self.__predicate__[type] = lambda v: fcompose(mapper, functools.partial(operator.eq, v))
+        self.__predicate__[type] = lambda target: fcompose(mapper, functools.partial(operator.eq, target))
```
arizvisa, line 446: This should probably be combinator based...but I really don't care to fuck with it right now.

base/_utils.py
```diff
@@ -432,21 +432,24 @@ class matcher(object):
...
     def boolean(self, type, function, *attribute):
         attr = self.__attrib__(*attribute)
-        self.__predicate__[type] = lambda v: fcompose(attr, functools.partial(function, v))
+        self.__predicate__[type] = lambda target: fcompose(attr, functools.partial(function, target))
```
arizvisa, line 449: This should probably be combinator based...but I really don't care to fuck with it right now.

base/_utils.py
```diff
@@ -432,21 +432,24 @@ class matcher(object):
...
-    def __attrib__(self, *attribute):
-        if not attribute:
-            return lambda n: n
-        res = [(operator.attrgetter(a) if isinstance(a, six.string_types) else a) for a in attribute]
-        return lambda o: tuple(x(o) for x in res) if len(res) > 1 else res[0](o)
+    def __attrib__(self, *attributes):
+        if not attributes:
+            return lambda item: item
+        res = [(operator.attrgetter(callable_or_attribute) if isinstance(callable_or_attribute, six.string_types) else callable_or_attribute) for callable_or_attribute in attributes]
+        return fcompose(*res) if len(res) > 0 else res[0]
```
arizvisa, line 439: This should be `1` not `0`.

arizvisa (commented):
Needs a single-character fix.

arizvisa (commented):
Okay. Maybe I'll test this prior to merging...

arizvisa:
Merging this thing after having to fix some bugs such as the formatting when emitting a structure member, and searching through comments with a glob when the `member_t.comment` attribute returns `None`.

Issue #114, opened 2021-01-17
Mapping keys with `ui.keyboard.map` can raise a `DisassemblerError` exception when mapping a key combination w/ a modifier


arizvisa:
When using the `ui.keyboard.map` function to map a function to a key, occasionally the `DisassemblerError` exception will get raised. This results in the inability to remap keys to a user-specified macro. This probably also happens with `ui.keyboard.unmap` because the underlying issue is with the `ui.keyboard.__of_key__` function.

For the record, this only seems to happen when using Python3 under PR #84. I'm pretty sure this is happening because sets on Py3 are ordered differently than on Py2.

arizvisa:
Linking this to PR #115 as it's responsible for fixing this issue.

PR #115, opened 2021-01-17
Implemented the `ui.keyboard.list` function and added better support for badly formatted input when using `ui.keyboard.map`.

arizvisa:
When mapping keys via the `ui.keyboard` namespace, there isn't any way to display the currently mapped keys and thus it is up to the user to determine what is mapped versus unmapped. When using hundreds of macros (like I apparently do), this very quickly becomes unmanageable. This PR implements a `ui.keyboard.list` function which does its best to display the currently managed key combinations along with the type of the callable that the combination is mapped to.

This PR originated due to an issue which manifested itself when using Python3 with PR #84 that was due to the `ui.keyboard.__of_key__` function mistakenly using the '_' character as a separator which isn't valid for the `idaapi.add_hotkey` function. To remedy this, this separator was simply removed and the input for the `ui.keyboard.__normalize_key__` was made to be more forgiving when given the hotkey to remap. This was done by allowing the user to specify their keyboard mapping via a string (i.e. "ctrl+f"), a tuple in the correct form (i.e. `(('ctrl',), 'f')`), or a set/list/tuple that mixes up the modifiers and key combinations (i.e. `{'ctrl','f'}`, `('ctrl','f')`, or `['ctrl','alt','f']`).

A conditional was also added to the `ui.keyboard.map` function to ensure that a valid callable is passed. This allows reducing the amount of type-checking that is done when performing a `ui.keyboard.list` of all the current keyboard mappings.

This fixes issue #114.

### DIFF
--- a/misc/ui.py
+++ b/misc/ui.py
@@ -914,6 +914,11 @@ class keyboard(object):
         else:
             res = None
 
+        # Verify that the user gave us a callable to use to avoid mapping a
+        # useless type to the specified keyboard combination.
+        if not builtins.callable(callable):
+            raise internal.exceptions.InvalidTypeOrValueError(u"{:s}.map({!s}, {!r}) : Unable to map the non-callable value {!r} to the hotkey combination {!s}.".format('.'.join([__name__, cls.__name__]), internal.utils.string.repr(key), callable, callable, internal.utils.string.repr(keystring)))
+
         # Define a closure that calls the user's callable as it seems that IDA's
         # hotkey functionality doesn't deal too well when the same callable is
         # mapped to different hotkeys.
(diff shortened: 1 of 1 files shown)


misc/ui.py
```diff
@@ -846,6 +926,11 @@ def map(cls, key, callable):
         else:
             res = None
 
+        # Verify that the user gave us a callable to use to avoid mapping a
```
arizvisa, line 929: This needs to be documented in the PR.

arizvisa (commented):
This requires editing the PR summary.

arizvisa:
Merging this fucker.

Issue #117, opened 2021-01-26
Function tags pointing to the wrong function chunk when enumerating them?


arizvisa:
In some cases, storing a global tag to a function will be cached with the wrong address rather than the actual function address. Instead of caching the function's address, it will cache the SEH handler (a different function chunk) for the function.

This manifests itself when enumerating all of the "database" tags in the database by reporting that the same queried tags are assigned to both the top of a function's chunk, and the top of the actual function (where the tags are actually located at). Essentially all consumers of the tag cache are reporting the address of the queried tags twice when enumerating them thus returning the same tags (but for different addresses) when fetching them thus confusing things like name information, counting, or similar.

arizvisa:
I'm theorizing that this is happening because of either some misuse of `function.top` or `function.address` in the key-determining function of the `internal.comments` module, a misimplementation of that mentioned function, or some hook is not promoting the function tag properly when IDA appends the chunk to the `idaapi.func_t`.

arizvisa:
Traceback 1:
```
  File "/home/user/work/ida-minsc/base/_interface.py", line 521, in closure
    callable(*args, **kwargs)
  File "/home/user/work/ida-minsc/base/_interface.py", line 382, in closure
    result = callable(*parameters)
  File "/home/user/work/ida-minsc/misc/hooks.py", line 738, in rename
    ctx.inc(ea, '__name__')
  File "/home/user/work/ida-minsc/base/_comment.py", line 1014, in inc
    traceback.print_stack()
CRITICAL:root:internal.interface.priorityhook.callback(269440528, 'SEH_10017EE0') : Callback for "IDP_Hooks.ev_rename" with priority (+0) raised an exception while executing <function rename at 0x7f025303e1d0>
WARNING:root:internal.interface.priorityhook.callback(269440528, 'SEH_10017EE0') : Traceback ("IDP_Hooks.ev_rename" was hooked at)
WARNING:root:internal.interface.priorityhook.callback(269440528, 'SEH_10017EE0') :   File "/home/user/idapro-7.5/python/2/ida_idp.py", line 6000, in dispatch
WARNING:root:internal.interface.priorityhook.callback(269440528, 'SEH_10017EE0') :     cb.fun(slot, *args)
WARNING:root:internal.interface.priorityhook.callback(269440528, 'SEH_10017EE0') :   File "/home/user/work/ida-minsc/base/_interface.py", line 382, in closure
[...]

PR #118, opened 2021-02-04
Removed the checks against the current analysis pass from all function-related hooks so that they always execute and update their cache.

arizvisa:
When a database is being analyzed, the plugin hooks a variety of points in order to determine the address and comments that need to be cached. In order to improve performance, the plugin keeps track of the current analysis state within a global variable named `State` and uses it to determine whether certain hooks should be executed or not. This was done so that we could avoid duplication of work as some of the reference counting is re-performed when the entire function is being processed.

During the different stages of IDA's analysis, a function may go through various states of promotion and demotion as analysis is being performed. As the plugin is hooking various parts of this analysis, this would occasionally result in the plugin performing actions that are later undone by analysis. To work around this, the plugin would try to determine whether IDA had performed its first analysis yet, and avoid performing any real cache-updating work until the first analysis has been completed. This way no serious python has been executed until all of the functions have been processed and cached.

It was discovered that occasionally a reference count would end up being created for a function at the address of an SEH handler, or an arbitrary function chunk which would result in the address of the chunk being returned instead of the function when querying the database. Examining the cache showed that there was commonly a single reference existing at the wrong address which would affect querying. This was happening because of the different passes that IDA performs on a database. Many of these wrong addresses were actually being cached because of the `__name__` tag. During an earlier pass, the suspect address would be created as a function, and then in a later pass the address would either be demoted into a chunk for another function, or the address would be removed due to an analysis pass changing the boundaries of the chunk. This would result in a stray reference to the `__name__` tag still being cached.

Although it was intended that these tag references be updated during the process of executing a function, it turned out that the reason why the stray reference was left was because the name at the given address was transformed to a global and then back to a function later during analysis which would leave the name in the cache. The hook wasn't executing because it was only intended to execute once the database has completed its first analysis pass.

To fix this, all of the relevant hooks were modified to remove the block which checks whether the database has completed its analysis or not. This can have an affect on performance as now during IDA's analysis process python may execute, but with my tests it doesn't seem to be as big of a deal as it was when I first wrote these hooks. I imagine back then that the bugs in the hooks (that have since been fixed) were affecting performance, and that was why the block was originally necessary.

This fixes issue #117.

### DIFF
--- a/misc/hooks.py
+++ b/misc/hooks.py
@@ -779,8 +779,6 @@ def func_tail_appended(pfn, tail):
     We simply iterate through the new chunk, decrease all of its tags in the
     global context, and increase their reference within the function context.
     """
-    global State
-    if State != state.ready: return
     # tail = func_t
     for ea in database.address.iterate(interface.range.bounds(tail)):
         for k in database.tag(ea):
@@ -796,8 +794,6 @@ def removing_func_tail(pfn, tail):
     We simply iterate through the old chunk, decrease all of its tags in the
     function context, and increase their reference within the global context.
     """
-    global State
-    if State != state.ready: return
     # tail = range_t
     for ea in database.address.iterate(interface.range.bounds(tail)):
         for k in database.tag(ea):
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
Pretty basic stuff. Hopefully performance isn't too bad when enabling these during analysis.

arizvisa:
Merging this thing.

Issue #120, opened 2021-02-15
ida 6.8 uses db.get.string() to get an error:


gool123456:
Hi, @arizvisa , ida 6.8 uses db.get.string() to get an error:

Python>db.get.string()
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 341, in database.string
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 5968, in string
    return cls.string(ui.current.address(), **length)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 341, in database.string
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 6084, in string
    res = cls.array(ea + shift, **length).tostring()
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 341, in database.array
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 5933, in array
    return decode_array(ea, T, length['length'], numerics, lnumerics)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\database.py", line 5849, in decode_array
    if _instruction.ops_refinfo(ea) or F & idaapi.FF_SIGN:
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 341, in instruction.ops_refinfo
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\instruction.py", line 1086, in ops_refinfo
    return op_refinfo(ea, OPND_ALL)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_utils.py", line 341, in instruction.op_refinfo
    return f(*arguments, **keywords)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\instruction.py", line 1095, in op_refinfo
    return ri if idaapi.get_refinfo(ri, ea, opnum) else None
[...]

PR #121, opened 2021-02-16
Fixed the parameter order for the `instruction.op_refinfo` function so that it's compatible with earlier versions of IDA.

arizvisa:
It appears that the order for the `idaapi.get_refinfo` function was changed in IDA 7.x. Prior to 7.x, the `idaapi.get_refinfo` function had the prototype of `idaapi.get_refinfo(ea_t, n, refinfo_t)` which was later changed to `idaapi.get_refinfo(refinfo_t, ea_t, n)`. This PR fixes the issue by explicitly checking the version and using it to distinguish whether to use the older parameter order or the newer one.

This fixes issue #120.


arizvisa:
The original reporter claimed that this fixes the issue. Gonna do a review, and then wait for a confirmation from him prior to merging.

arizvisa (commented):
looks okay...

arizvisa:
> The original reporter claimed that this fixes the issue. Gonna do a review, and then wait for a confirmation from him prior to merging.

Original reporter just confirmed. Merging...

Issue #123, opened 2021-04-08
Displaying the hooks and notifications that are currently active


arizvisa:
It would be pretty nice to display the hooks and notifications that are currently active on the objects. This might make the `ui.hook` objects a lot more approachable as right now a user would need to either examine the object themselves or review the source to see the priorities that are currently being used. Plus, enabling and disabling targets is pretty non-intuitive.

arizvisa:
Just push'd the branch, now I need to create the PR...

arizvisa:
This issue is fixed by PR #124.

arizvisa:
Closing this issue as PR #124 was merged.

PR #124, opened 2021-04-08
Updated the hook and notification objects to render the targets that are currently being used, and whether they're enabled or disabled.

arizvisa:
This PR updates the `prioritybase` class in the `internal.interface` module so that it displays all of the targets that are hooked by the object, and displays each priority/callable pair that is associated with a given target. The enabled and disabled targets are emitted within their own sections, and the list of callables is sorted by priority.

To accomplish this, the `prioritybase` class had a few new properties added so that a user can ask the object what targets are enabled or disabled. Also, the `prioritynotification` class also had its formatter updated so that the notification targets are rendered according to the `idaapi.NW_*` enumeration from IDAPython.

During the implementation of this, a minor issue was discovered when removing the last target while it's disabled. This would result in the target not being completely removed from the cache. This was patched by ensuring that when the target is removed, any reference to the target in the disabled set is also removed.

This closes issue #123.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -329,7 +329,7 @@ class prioritybase(object):
             # If a method is passed to us, then we need to extract all
             # of the relevant components that describe it.
             if isinstance(object, (types.MethodType, staticmethod, classmethod)):
-                cls = pycompat.method.type(object)
+                cls = pycompat.method.self(object)
                 func = pycompat.method.function(object)
                 module, name = func.__module__, pycompat.function.name(func)
                 iterable = parameters(func)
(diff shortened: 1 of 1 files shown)


arizvisa:
Force-push'd due to rebase.

base/_interface.py
```diff
@@ -369,6 +454,7 @@ def discard(self, target, callable):
         # from our cache.
         else:
             self.__cache__.pop(target, [])
+            self.__disabled.discard(target)
```
arizvisa, line 457: This line is a bugfix.

arizvisa (commented):
Just one thing to update the conversation with.

arizvisa:
Okay. Finished updating the PR description. Just need to link this, and then merge it.

Issue #122, opened 2021-04-08
When building a database the names for some of the addresses are not cached in the tag cache


arizvisa:
As reported by a user, in some cases when building the tag cache from a fresh database, a new `__name__` attribute does not get its reference tracked in the tag cache. This was actually done by design so that IDA's default names that it assigns during analysis are ignored when querying a database. This way querying things with `database.select` will only select things that you've actually touched. That's why things like extra comments, names, and type information have their tags double-underscored (`__extra_prefix__` and `__extra_suffix__`, `__name__`, and `__typeinfo__`).

Despite this being intentional, this is not consistent with how people expect tags to work since getting the tag for an address will always return these dictionary keys with their values. The `custom.tagfix` module actually indexes __everything__, so as a result the history that I'm trying to preserve is explicitly lost when rebuilding the cache.

arizvisa:
I'll need to figure out whether I should fix this so that it's always consistent, or I should keep the original logic which is apparently only useful to me. Originally the idea was that anything which is double-underscored is an implicit tag and is never indexed unless it's being explicitly set by the user. This is true for most things...except for the `__color__` tag (because I forgot about it and never use it), and the `__typeinfo__` tag (because I didn't think those hooks actually worked).

At the very least, this characteristic should be documented so that people who're actually using the tagging interface have knowledge about it. Some research should also be done to see if we can hook when the `__typeinfo__` for an address and `__color__` has changed so that we can update the cache when a user touches it.

arizvisa:
I guess I'm going to create a separate branch for making it consistent, and then test whether it makes sense to retain the original functionality or to just index everything.

arizvisa:
PR #125 is an implementation where the cache keeps track of all implicit tags during analysis and tagfix. Personally I don't like it...

arizvisa:
Okay. So if we keep track of some of the implicit tags the user may modify, then we can use these hooks.

Used when changing the function prototype or the type for a data item:
```
# changing_ti(ea, new_type, new_fnames)
ui.hook.idb.add('changing_ti', Fchanging_ti, 1000)
# ti_changed(ea, type, fnames)
ui.hook.idb.add('ti_changed', Fchanged_ti, 1000)
```
Used when changing an instruction operand type:
```
# changing_op_type(ea, n, opinfo)
ui.hook.idb.add('changing_op_type', Fchanging_opt, 1000)
# op_type_changed(ea, n)
[...]

PR #125, opened 2021-04-08
Updated both the hooks and tagfix modules to update the tag cache for any and all types of tags

arizvisa:
This PR modifies the `custom.tagfix` and `hooks` modules to ensure that all implicit tags are being written to the tag cache.

Specifically, the `hooks` module will now treat the database being initialized as it being ready which results in any name changes or comments that are made by IDA when it's performing its analysis being written to the tag cache. This might affect performance slightly, but it's unnoticable on my box with reasonably sized databases (but I'm not testing it against something like acrord or mso). The `custom.tagfix` module has also been specifically updated to crawl the `__extra_prefix__`, `__color__`, and `__name__` tags when processing both function contents and globals.

This results in the `hooks` grabbing almost all of the tags, but misses the "extra" comments that are represented by the `__extra_prefix__` and `__extra_suffix__` tags. I didn't spend too much time on this, but I figure that these aren't being identified correctly because Ilfak didn't describe the "extra" comment hooks with the same semantics as the regular comments ("comment changing" versus "comment changed").

This fix is one of the proposed solutions to issue #122.

### DIFF
--- a/custom/tagfix.py
+++ b/custom/tagfix.py
@@ -49,6 +49,30 @@ def fetch_contents(fn):
             for name in items:
                 addr[ea] = addr.get(ea, 0) + 1
                 tags[name] = tags.get(name, 0) + 1
+
+            # check if there's any extra comments so we can explicitly add those too.
+            eprefix = db.extra.__get_prefix__(ea)
+            if eprefix is not None:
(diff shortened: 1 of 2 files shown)


arizvisa:
Closing this PR as it is being dropped in favor of PR #126.

PR #126, opened 2021-04-08
Added more hooks in order to support the inclusion of more implicit tags in the tag cache, and added a workaround for the situation where a "changing_ti" event is dispatched without its related "ti_changed" event properly following it.

arizvisa:
This PR adds more hooks in order to catch when more of the hidden tags are modified by a user. This includes colors (`__color__`), and type info (`__typeinfo__`). When any of these are changed by a user, the tag cache will be updated with the user's changes. This should allow more ways for a user to query the parts of a database that they've visited before.

The hook to track the "extra" comments tags, `__extra_prefix__` and `__extra_suffix__`, has been reimplemented so that it's not a noop. Instead of tracking each individual extra comment index, it was rewritten to only check if the base index (`idaapi.E_PREV` or `idaapi.E_NEXT`) exists or not. This way we can guarantee that the tag is removed from the cache when the comment itself is removed. This unfortunately means that when "extra" comments are updated (as opposed to added or removed), and the modification has no change in the number of lines, we end up losing track of the refcount, and its address may remain permanently in the cache.

While testing this, I encountered another issue related to the "changing_ti" event being sent without its "ti_changed" partner. I was able to figure out a workaround that seems to work for most situations, but it's definitely pretty hacky. I'm considering refactoring this at the cost of potential memory loss, but I'll submit that in a separate PR.

This is one of the other solutions to issue #122.

### DIFF
--- a/misc/hooks.py
+++ b/misc/hooks.py
@@ -769,6 +769,34 @@ def extra_cmt_changed(ea, line_idx, cmt):
         continue
     return
 
+### individual tags
+def item_color_changed(ea, color):
+    '''This hook is for when a color is applied to an address.'''
+
+    # First make sure it's not an identifier, as if it is then we
+    # need to terminate early because the tag cache doesn't care
+    # about this stuff.
+    if interface.node.is_identifier(ea):
+        return
+
+    # Now we need to distinguish between a content or global tag so
+    # that we can look it up to see if we need to remove it or add it.
+    ctx = internal.comment.contents if idaapi.get_func(ea) else internal.comment.globals
+
+    # FIXME: we need to figure out if the color is being changed,
+    #        updated, or removed. since there's no way to determine
+    #        this accurately, we just assume that any color is going
+    #        to increase the reference count.
+
+    # If the color was restored, then we need to decrease its ref.
+    if color in {idaapi.COLOR_DEFAULT}:
+        ctx.dec(ea, '__color__')
+
+    # The color is being applied, so we can just increase its reference.
+    else:
+        ctx.inc(ea, '__color__')
+    return
+
 ### function scope
 def thunk_func_created(pfn):
     pass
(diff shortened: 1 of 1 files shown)


misc/hooks.py
```diff
@@ -1080,13 +1305,29 @@ def make_ida_not_suck_cocks(nw_code):
...
+
     ## just some debugging notification hooks
     #[ ui.hook.ui.add(item, notify(item), -100) for item in ['range','idcstop','idcstart','suspend','resume','term','ready_to_run'] ]
     #[ ui.hook.idp.add(item, notify(item), -100) for item in ['ev_newfile','ev_oldfile','ev_init','ev_term','ev_newprc','ev_newasm','ev_auto_queue_empty'] ]
     #[ ui.hook.idb.add(item, notify(item), -100) for item in ['closebase','savebase','loader_finished', 'auto_empty', 'thunk_func_created','func_tail_appended'] ]
     #[ ui.hook.idp.add(item, notify(item), -100) for item in ['add_func','del_func','set_func_start','set_func_end'] ]
     #ui.hook.idb.add('allsegs_moved', notify('allsegs_moved'), -100)
     #[ ui.hook.idb.add(item, notify(item), -100) for item in ['cmt_changed', 'changing_cmt', 'range_cmt_changed', 'changing_range_cmt'] ]
+    #[ ui.hook.idb.add(item, notify(item), -100) for item in ['changing_ti', 'ti_changed', 'changing_op_type', 'op_type_changed'] ]
+    #[ ui.hook.idb.add(item, notify(item), -100op_type_changed for item in ['changing_op_ti', 'op_ti_changed'] ]
```
arizvisa, line 1330: This commented line includes a type-o and is also missing some other events that can use notify.

arizvisa (commented):
Minor fix to a comment that appears to suffer from a mis-pasted register.

arizvisa (commented):
Looks good.

arizvisa:
Testing a build, and then will maybe merge once it works.

arizvisa:
Found a bug...of course.

arizvisa:
Force push'd due to the desire to reword a commit.

arizvisa (commented):
I am tired, but this looks okay.

arizvisa:
Merging this fucker.

Issue #127, opened 2021-04-09
Some of the hooks can raise exceptions due to IDA dispatching "changed" events out-of-order.


arizvisa:
Many of the hooks are actually pairs of events that are supposed to be dispatched to an implementor. These use a "changing" event which is then followed by a "changed" event. The "changing" events are responsible for informing an implementor what the original value is and then afterwards the "changed" event will inform the implementor that the change has been written to the database.

In some instances the "changed" even might never appear. This used to occur because of bad code raising an exception inside a coroutine, but apparently the "changing_ti" and "ti_changed" events can do this after IDA has completed its analysis. In PR #126, I implemented a workaround in the `hooks.typeinfo` namespace by using an exception and looping until the right even has been received, but this is a pretty terrible hack.

I think it'd be significantly more effective if we trade some memory away to store a coroutine for every single address we receive the "changing" event on, and then use the "changed" event to fetch and complete the coroutine. This can likely be done for each of the events that I'm hooking such as for comments and area comments.

arizvisa:
This issue is fixed by PR #128.

arizvisa:
PR #128 was merged. Closing this issue.

PR #128, opened 2021-04-10
Refactored all of the hooks based on "changing" and "changed" events to store and resume their state by their address.

arizvisa:
When handling certain events such as the `IDB_Hooks.changing_ti` and `IDB_Hooks.ti_changed` hooks, IDA doesn't guarantee that they will be called in order. This affected the previous way the hooks were being executed because they were assuming that every "changing" hook would be followed by a "changed" one. The previous PR, #126, added support for updating the tag cache with any changes to type information. After IDA has analyzed a database, updates to type information were not guaranteed to call the "changed" hook which resulted in the implementation of a hack to deal with out-of-order events being dispatched.

This PR fixes the issue by refactoring the logic for an event, and instead constructing an individual throwaway coroutine whenever a "changing" event was received. These coroutines are then keyed by address and stored in a dictionary so that way if the corresponding "changed" event is not dispatched, then the coroutine can just be resetted. Each of the hooks that follow the "changing" and "changed" methodology were updated to use the implementation introduced by this PR.  

This resolves issue #127.

### DIFF
--- a/misc/hooks.py
+++ b/misc/hooks.py
@@ -248,11 +248,10 @@ class address(changingchanged):
         try:
             event.send((ea, bool(repeatable_cmt), utils.string.of(newcmt)))
 
-        # If a StopIteration was raised when submitting the comment to the
-        # coroutine, then we somehow desynchronized. Re-initialize the coroutine
-        # with the hope that things are fixed.
+        # If a StopIteration was raised when submitting the comment to the coroutine,
+        # then something failed and we need to let the user know about it.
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
I'm tired, but I \_did\_ check all the formatspecs used by the log records and they looked okay.

arizvisa:
Merging this. This method of dealing with mismatched hooks should be significantly more reliable.

Issue #129, opened 2021-04-12
Some calculations used when calculating the number of digits at various bases are incorrect.


arizvisa:
Various places attempt to calculate the number of digits of a binary, decimal, and a hexadecimal number using logarithms. This should typically be `f(number, base) = 1 + ⌊log(number, base)⌋`. However, due to python imprecision this can be incorrect.

An example of where this will not emit the correct result is when the value approaches 1000. With python, the result for `log(1000, 10)` is `math.log(1000, 10) = 2.9999999999999996`. The previous integer, 999, is `math.log(999, 10) = 2.9995654882259823`. Thus if we take the floor of this value, it'll take just the integer which is incorrect. If we add `1e-15` to it, then that might fix it for powers of 10, but powers of 2 or 16 will still need to be calculated differently.

arizvisa:
This issue is fixed by PR #130.

arizvisa:
Closing this issue as PR #130 was just merged.

PR #130, opened 2021-04-12
Fixed all the precision issues with any logarithm calculations that are used to determine the number of digits for various bases.

arizvisa:
Due to python's imprecision, certain logarithms will result in the wrong value where the value is off by around `1e-15`. Due to this, taking the floor of the logarithm's value will result in a completely wrong result. This type of calculation is used in a variety of places in order to determine the number of bits or digits for a given integer. Some examples where the logarithm result is off is when the base 10 logarithm is calculated for 1000. Logarithms for base 2 will also demonstrate an incorrect result for similar reasons.

This PR attempts to fix these calculations by ensuring that `math.log10` is used for any base 10 calculations, and when calculating base 2 calculations the ceiling is taken rather than the floor. Testing these changes against a number of boundaries for powers of 2 seems to result in the "length" of the integer being of the expected value.

We now ensure that base 10 numbers are always calculated with `f(n) = 1 + math.floor(math.log10(n))` which uses the `math.log10` function available in both Python2 and Python3 and demonstrates correctness. As the `math.log2` function is only available on Python3 (and not Python2), the function `f(n, b) = math.ceil(math.log(n, b))` is used for calculating the logarithm. WIth this, base 2 numbers use `f(n, 2) = math.ceil(math.log(n, 2))`, and base 16 numbers  use `f(n, 16) = math.ceil(math.log(n, 16))`.

This closes issue #129.

### DIFF
--- a/base/segment.py
+++ b/base/segment.py
@@ -103,10 +103,12 @@ def list(**type):
 
         listable.append(seg)
 
-    # Collect the maximum sizes for everything from the first pass
-    Fdigits = lambda number, base: math.floor(1. + math.log(number or 1) / math.log(base))
-    cindex = Fdigits(maxindex, 10)
-    caddr, csize = (Fdigits(item, 16) for item in [maxaddr, maxsize])
+    # Collect the maximum sizes for everything from the first pass. We have
+    # to use different algorithms as due to Python's issues with imprecision,
+    # the resulting number of digits will vary depending on what base is
+    # actually being used when calculating the logarithm.
+    cindex = 1 + math.floor(math.log10(maxindex or 1))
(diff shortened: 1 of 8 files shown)


base/segment.py
```diff
@@ -165,7 +167,7 @@ def by(**type):
     listable = [item for item in __iterate__(**type)]
     if len(listable) > 1:
         maxaddr = max(builtins.map(interface.range.end, listable) if listable else [1])
-        caddr = math.floor(1. + math.log(maxaddr or 1) / math.log(16))
+        caddr = math.ceil(math.log(maxaddr or 1, 16))   # only works with powers of 16
```
arizvisa, line 170: not too important, but this should mention that it only works with powers of 2 and related.

arizvisa (commented):
A comment might need to be updated, but it's not relevant at all.

arizvisa:
LGTM...merging this shiz.

PR #131, opened 2021-05-06
Improved structure support for dealing with unions, and fixed a few bugs that were related.

arizvisa:
This PR modifies the classes defined by the structure module in order to allow proper support of unions. This involved fixing and reworking a few methods so that it can still be used in mostly the same way. As a side-effect of this, the documentation was also significantly improved. This code was originally from like a decade ago, and so some things which have changed since then have been updated.

The major rewrite has been the `members_t.by_realoffset`, and `members_t.__walk_to_realoffset__` methods. In most cases, when a structure is a union the implementation will try its harded to narrow down the union's members to the offset the user is looking for. If the exact member isn't found, then the resulting members will be filtered in order to find the exact one. In the case of the `members_t.by_realoffset` method, the implementation will try and find the "exact" field that is being pointed to by sorting the resulting member list by giving priority to members where the provided offset is pointing to the boundaries inside a member. The `members_t.__walk_to_realoffset__` method allows the user to filter the members explicitly by providing a closure that can be used to filter the results.

This aims to add the feature's requested by issue #6.

### DIFF
--- a/base/structure.py
+++ b/base/structure.py
@@ -1472,10 +1472,9 @@ class members_t(object):
             prefix = (self.by_identifier(item.id) for item in [mptr])
             return builtins.tuple(prefix), offset - moffset
 
-        # Otherwise, the member type's a structure, so we need to
-        # recurse to figure out which field should be at the relative
-        # offset. Afterwards, we then shift the relative offset back
-        # into a real offset so that we can collect it in the result.
+        # Otherwise, the member type is a structure, and we'll need
+        # to recurse in order to figure out which field should be at
+        # the relative offset from the member.
         st = __instance__(sptr.id, offset=self.baseoffset + moffset)
         result, nextoffset = st.members.__walk_to_realoffset__(offset - moffset, filter=filter)
 
@@ -1496,7 +1495,7 @@ class members_t(object):
 
         # Now we can concatenate our prefix to our current results,
         # and then return what we've aggregated back to our caller.
-        return prefix + result, offset - (moffset + nextoffset)
+        return prefix + result, nextoffset
 
     def near_offset(self, offset):
         '''Return the member nearest to the specified `offset` from the base offset of the structure.'''
--- a/base/structure.py
+++ b/base/structure.py
@@ -226,6 +226,19 @@ by_id = byidentifier = byId = utils.alias(by_identifier)
 ## FIXME: need to add support for a union_t. add_struc takes another parameter
 ##        that defines whether a structure is a union or not.
 
+@utils.multicase(id=six.integer_types)
+def is_union(id):
+    '''Return whether the structure identified by `id` is a union or not.'''
+    structure = idaapi.get_struc(id)
+    return is_union(id)
+@utils.multicase()
+def is_union(structure):
+    '''Return whether the provided `structure` is defined as a union.'''
+    if isinstance(structure, structure_t):
+        return is_union(structure.ptr)
+    return structure.is_union()
+unionQ = isunion = utils.alias(is_union)
+
 ### structure_t abstraction
 class structure_t(object):
     """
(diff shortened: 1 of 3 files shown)


base/structure.py
```diff
@@ -1112,63 +1170,188 @@ def by_offset(self, offset):
...
 
         # Guard against a potential OverflowError that would be raised by SWIG's typechecking
-        if self.baseoffset > maximum:
-            cls = self.__class__
-            raise E.MemberNotFoundError(u"{:s}({:#x}).members.by_offset({:+#x}) : Unable to find member at specified offset ({:+#x}).".format('.'.join([__name__, cls.__name__]), owner.id, offset, offset))
-
-        # Look for the very last member so we can confirm our bounds.
-        mptr = idaapi.get_member(owner.ptr, maximum - self.baseoffset)
-        if mptr is None:
+        if not (minimum <= self.baseoffset < maximum):
```
arizvisa, line 1176: this is very wrong and should be "offset" instead of "self.baseoffset".

base/structure.py
```diff
@@ -1112,63 +1170,188 @@ def by_offset(self, offset):
...
+    def __members_at__(self, realoffset):
+        """Yield all the members at the specified `realoffset` of the current structure.
+
+        This returns members whilst keeping in mind whether the structure is a union and may have more than one field at the same offset.
+        """
+        owner = self.owner
+
+        # If this is not a union, then this is simple because there's only one member
+        # at any given offset. IDAPython's api seems to also figure everything out for
+        # us and so we can just yield things and just return.
```
arizvisa, line 1193: Your english sucks, Ali.

base/structure.py
```diff
@@ -1112,63 +1170,188 @@ def by_offset(self, offset):
...
-        mem = idaapi.get_member(owner.ptr, offset)
-        if mem is None:
+        # If we found more than one result, then we need to warn the user about it
+        # because we're going to have to make a decision on their behalf. This really
+        # should only be happening when we're a union type.
+        if len(items) > 1:
             cls = self.__class__
-            raise E.MemberNotFoundError(u"{:s}({:#x}).members.by_realoffset({:+#x}) : Unable to find member at specified offset ({:+#x}).".format('.'.join([__name__, cls.__name__]), owner.id, offset, offset))
+            iterable = (idaapi.get_member_fullname(mptr.id) for mptr in items)
+            logging.warning(u"{:s}({:#x}).members.by_realoffset({:+#x}) : The requested offset ({:#x}) contains more than one member ({:s}).".format('.'.join([__name__, cls.__name__]), owner.id, offset, offset, ', '.join(map(utils.string.to, iterable))))
```
arizvisa, line 1265: Maybe change the warning so it reads as "The specified offset ({:#x}) is occupied by more than one member ({:s})".

base/structure.py
```diff
@@ -1112,63 +1170,188 @@ def by_offset(self, offset):
...
+
+            # Grab the type information for each member so we can determine if the
+            # requested offset points at an array or a structure.
+            candidates = []
+            for mptr in items:
+                opinfo = idaapi.opinfo_t()
+                res = idaapi.retrieve_member_info(mptr, opinfo) if idaapi.__version__ < 7.0 else idaapi.retrieve_member_info(opinfo, mptr)
```
arizvisa, line 1272: As the result of res is being explicitly used, add a comment describing how it's being used.

base/structure.py
```diff
@@ -1190,84 +1373,130 @@ def by_identifier(self, id):
...
-            return [m] + [item for item in res], arrayoffset - (memberoffset - nextoffset)
-
-        # Any other type means we can just add it as an offset since we don't
-        # have any other way of representing these things.
-        return [m], arrayoffset
+        If a closure is passed as the `filter` parameter, then use the function to filter the members to use when descending into a structure.
+        """
+        owner = self.owner
+
+        # Grab the type information for a member, and convert it to
```
arizvisa, line 1383: Note that we're defining a closure that does this.

base/structure.py
```diff
@@ -1276,36 +1505,39 @@ def near_realoffset(self, offset):
...
             cls = self.__class__
-            logging.info(u"{:s}({:#x}).members.near_realoffset({:+#x}) : Requested offset not within bounds {:#x}<->{:#x}. Trying anyways..".format('.'.join([__name__, cls.__name__]), owner.id, offset, minimum, maximum))
-
-        # Try and find the exact offset, because if we can..then we don't need
-        # to execute the rest of this function.
-        mem = idaapi.get_member(owner.ptr, offset)
-        if mem:
-            index = self.index(mem)
-            return self[index]
+            logging.warning(u"{:s}({:#x}).members.near_realoffset({:+#x}) : Requested offset not within bounds {:#x}<->{:#x}. Trying anyways..".format('.'.join([__name__, cls.__name__]), owner.id, offset, minimum, maximum))
```
arizvisa, line 1511: This would read better as "Requested offset is not within the bounds of the structure ({:#x}<>{:#x})".

base/structure.py
```diff
@@ -1314,14 +1546,21 @@ def recurse(offset, available):
...
-        return self.add(name, int, offset)
+        return self.add(name, int)
     @utils.multicase(name=(six.string_types, tuple))
     @utils.string.decorate_arguments('name')
     def add(self, name, type):
         '''Append the specified member `name` with the given `type` at the end of the structure.'''
+
+        # If this structure is a union, then the offset is always 0.
+        if is_union(self.ptr):
+            return self.add(name, type, 0)
```
arizvisa, line 1557: This might need to be "self.baseoffset" instead of 0 due to the way it's used in the code that follows.

base/structure.py
```diff
@@ -1331,6 +1570,12 @@ def add(self, name, type, offset):
         """
         flag, typeid, nbytes = interface.typemap.resolve(type)
 
+        # Check if the user is trying to add a union, if so then we need to
```
arizvisa, line 1573: "...add _to_ a union..."

base/structure.py
```diff
@@ -1568,22 +1827,27 @@ def id(self):
...
     def offset(self):
         '''Return the offset of the member.'''
-        return self.ptr.get_soff() + self.parent.members.baseoffset
+        return self.realoffset + self.parent.members.baseoffset
     @property
     def flag(self):
         '''Return the "flag" attribute of the member.'''
-        m = idaapi.get_member(self.parent.ptr, self.offset - self.parent.members.baseoffset)
-        return 0 if m is None else m.flag
+        mptr = idaapi.get_member(self.parent.ptr, self.offset - self.parent.members.baseoffset)
```
arizvisa, line 1849: Assign the "parent" variable first.

arizvisa (commented):
A couple of fixes and missing comments to add.

arizvisa (commented):
I did review this already, and I'm not going to do it again.

arizvisa (commented):
This "changed from last view" thing in github really blows. Anyways, I'm marking this as reviewed because this took too damn long. I'll just fix it in post.

arizvisa:
Merging this. Omfg...hopefully I never have to do this ever again.

Issue #133, opened 2021-05-17
The `instruction.op_refs` function doesn't return all references when applied to structure paths that use unions.


arizvisa:
As a result of PR #131 which introduced support for unions, the `op_refs` function doesn't seem to return all of the operand references when that operand is pointing to a union member. After some digging, it turns out that this is because the `idaapi.get_stroff_path` doesn't always return the all the members in the structure path for a given operand.

Actually, what's really happening is that IDA doesn't store all of the members for a structure path within the given operand. I'm assuming that in order to save on space, IDA can only save a maximum of `idaapi.MAXSTRUCPATH` ids in the `NSUP_STROFFX` supval for an instruction. I found some docs that say only the union ids are stored, but I've also found addresses in my netbios.sys database where other IDs are stored as well. These inconsistencies are fucking killing me.

Anyways, it seems like we'll need to do the calculations ourselves. This will require re-working the `instruction.op_refs` function since it returns all references for globals, locals, references, and enumerations...fuck.

arizvisa:
This is fixed by PR #134....and btw, fuck IDA's `get_stroff_path` api. I can't believe I had to fucking do everything myself.

arizvisa:
Closing this issue due to the associated PR being merged.

PR #134, opened 2021-05-17
Fixed the implementation of the `instruction.op_refs` function so that it supports all of the available reference types.

arizvisa:
When the enhancement introduced by PR #131, the `instruction.op_refs` function would not reliably return all of the references for a given operand. This included enumerations and most importantly structure paths where a member in the path is an explicit union. Structure offset operands that did not include any members (such as size, or references to the structure) would also not be included in the returned results.

This PR fixes all of those issues, and reworks the `instruction.op_structure` function so that it includes support for structure paths that do not have any members. The `instruction.op_refs` function also supports these properly, and it also includes references to both globals and stack variables that use the operand specified by the user. The delta parameter was also removed from the `instruction.op_structure` function instead opting to include it via the path. The original cases for `instruction.op_structure` rip the delta out of the keyword and push it to the end of the path in order to remain compatible with the previously existing api.

This fixes issue #133.

### DIFF
--- a/base/instruction.py
+++ b/base/instruction.py
@@ -819,8 +819,47 @@ def op_structure(opnum):
 def op_structure(ea, opnum):
     '''Return the structure and members for the operand `opnum` at the instruction `ea`.'''
     F, op = database.type.flags(ea), operand(ea, opnum)
+
+    # First check if our operand is pointing to memory by checking the operand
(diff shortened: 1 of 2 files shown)


arizvisa (commented):
I'll fix this in post. Nobody cares anyways, and I just want to fucking finish reversing this driver.

Issue #135, opened 2021-05-27
On older versions of IDA, the `idaapi.get_member_by_id` function will return a malformed `idaapi.struc_t`.


arizvisa:
For some reason, the `idaapi.get_member_by_id` function from IDAPython can return an `idaapi.struc_t` that appears to be malformed. This function is \_supposed\_ to return a tuple composed of `(mptr: idaapi.member_t, fullname: str, sptr: idaapi.struc_t)`. However, in some cases I've been getting an `sptr` that returns a bunk value whenever I try to access its `id` property.

Typically the value it returns is zero, but here's an example where it returns 48.
```
Python>id=0xff000000000003f6L
Python>a,b,c=idaapi.get_member_by_id(id)
Python>c.id
1394703794176
Python>x=c.id
Python>x
48
```

arizvisa:
Tagging this as both a bug and an enhancement, because the fix for this will be a workaround for an IDAPython issue.

arizvisa:
Closing this issue as it was fixed by PR #136.

PR #136, opened 2021-05-27
Added guards around all instances of the `idaapi.get_member_by_id` function to guarantee a proper `idaapi.struc_t` is returned.

arizvisa:
The `idaapi.get_member_by_id` function is supposed to return a tuple composed of the `idaapi.member_t`, its full name, and the `idaapi.struc_t` that contains it. In some cases, this IDAPython api can return a busted-up `idaapi.struc_t`. This `idaapi.struc_t` acts pretty fucked up, and doesn't let you access its id properly which we completely depend on. This api is used by the `instruction.op_structure` set of functions and the `structure` module.

To work around this issue, we first attempt the original `idaapi.get_member_by_id` call. Once we unpack the result, we immediately check the id for the `idaapi.struc_t` that is returned. If this id is undefined, then we re-fetch the `idaapi.struc_t` using a combination of `idaapi.get_member_struc` and `idaapi.get_member_fullname` using the original `idaapi.member_t` identifier.

This adds the necessary workaround for resolving issue #135.

### DIFF
--- a/base/instruction.py
+++ b/base/instruction.py
@@ -877,6 +877,8 @@ def op_structure(ea, opnum):
         table = {}
         for mptr in members:
             _, fullname, sptr = idaapi.get_member_by_id(mptr.id)
+            if not interface.node.is_identifier(sptr.id):
+                sptr = idaapi.get_member_struc(idaapi.get_member_fullname(mptr.id))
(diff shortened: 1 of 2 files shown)


arizvisa (commented):
I can't believe this had to be patched. What the fuck is SWIG actually doing?

Issue #137, opened 2021-07-11
Missing "_ida_netnode.netnode_exist" function in IDA Pro 7.5.201028.


c0d3z3r0:
`netnode_exist` does not seem to exist in v7.5.201028.

~~~
CRITICAL:root:internal : Error trying to import module `netnode` from C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\ida_minsc\base\_netnode.py.
Traceback (most recent call last):
  File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\idapythonrc.py", line 148, in load_module
    res = self.new_api(name, path)
  File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\idapythonrc.py", line 125, in new_api
    res = super(cls, self).new_api(fullname, path)
  File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\idapythonrc.py", line 67, in new_api
    return self.imp.load_module(modulename, file, path, description)
  File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\ida_minsc\base\_netnode.py", line 17, in <module>
    class netnode(object):
  File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\ida_minsc\base\_netnode.py", line 42, in netnode
    exist, exist_name = _ida_netnode.exist, _ida_netnode.netnode_exist
AttributeError: 'module' object has no attribute 'netnode_exist'
~~~

Edit:

Interstingly various files match that pattern... wtf.

~~~
IDA SDK and Tools/idasdk75/module/java/npool.cpp:  if ( can_create && !netnode_exist(n) )
IDA SDK and Tools/idasdk75/module/java/npool.cpp:    QASSERT(10333, netnode_exist(n));
IDA SDK and Tools/idasdk75/include/netnode.hpp:idaman bool ida_export netnode_exist(const netnode &n);
IDA SDK and Tools/idasdk75/include/netnode.hpp:  friend bool exist(const netnode &n) { return netnode_exist(n); }
...
grep: python/2/ida_32/_ida_nalt.pyd: binary file matches
grep: python/2/ida_32/_ida_netnode.pyd: binary file matches
grep: python/2/ida_64/_ida_nalt.pyd: binary file matches
grep: python/2/ida_64/_ida_netnode.pyd: binary file matches
...
grep: ida64.dll: binary file matches
grep: ida64.exe: binary file matches
grep: idat64.exe: binary file matches
[...]

PR #138, opened 2021-07-12
Added a fallback implementation for instances of IDA where the _ida_netnode.netnode_exist isn't available.

arizvisa:
On earlier versions of IDA (5.6), the `netnode_exist` function is defined in the SDK, but is not actually available in IDAPython. As a result of this discovery, a fallback implementation was defined in commit ff3b1a907f4e05934cadfaece64a2c2c0ad65b70 that used the `_ida_netnode.new_netnode` function with its "do_create" parameter set to `False`. This implementation was then chosen when an earlier version of IDA was detected.

Unfortunately, as reported in issue #137, it was discovered that later versions of IDA don't always include this function in IDAPython. As a result of this, it was decided to replace the version-check that distinguishes whether to use the fallback implementation (or not) with an explicit check for the `_ida_netnode.netnode_exist` function. If this function exists within the `_ida_netnode` module, then we assign it to `netnode.exist_name`. If it doesn't, then we fallback to the `_ida_netnode.new_netnode` implementation to determine if a netnode with the given name is defined within the database.

This fixes issue #137.

### DIFF
--- a/base/_netnode.py
+++ b/base/_netnode.py
@@ -36,12 +36,16 @@ class netnode(object):
     index = _ida_netnode.netnode_index
     kill = _ida_netnode.netnode_kill
 
-    # these apis exist in 5.6 of the sdk, but there's
-    # definitely a chance that they're not in IDAPython.
-    if idaapi.__version__ > 7.1:
-        exist, exist_name = _ida_netnode.exist, _ida_netnode.netnode_exist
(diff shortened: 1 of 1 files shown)


c0d3z3r0 (approved):
Tested successfully. Thank you!

arizvisa:
Merging as the PR was confirmed by the OP.

Issue #139, opened 2021-07-12
The `enumeration.has` function can return `False` when given a valid enumeration identifier.


arizvisa:
The following was reported in issue #137 while the original reporter was troubleshooting the issue. The `enumeration.by_identifier` function uses the `enumeration.has` implementation to determine whether the given identifier has an enumeration mapped to it. The implementation perfmrs this check by verifying that the provided identifier is a valid id for the database type (masked with 0xFF000000 for 32-bit, and 0xFF00000000000000 for 64-bit) and then checking if the relevant enumeration "altvals" are defined for it. This low-level way of checking if something is an enumeration was done because on earlier versions of IDA, IDA would consider some "custom data" or "structure" identifiers as enumerations which is wrong.

---
hm, this seems to have to do with netnode, too, since it uses `has`:

~~~
Python>enum.by_identifier(0xff00020f)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\ida_minsc\base\enumeration.py", line 114, in by_identifier
    raise E.EnumerationNotFoundError(u"{:s}.by_identifier({!s}) : Unable to locate the enumeration with the specified identifier ({:#x}).".format(__name__, eid, eid))
internal.exceptions.EnumerationNotFoundError: enumeration.by_identifier(4278190607) : Unable to locate the enumeration with the specified identifier (0xff00020f).
Python>idaapi.get_enum_name(0xff00020f)
'RAM1_01'
~~~

_Originally posted by @c0d3z3r0 in https://github.com/arizvisa/ida-minsc/issues/137#issuecomment-877783107_

arizvisa:
Ftr, this issue was reported in IDA v7.5.201028 using a 32-bit database.

arizvisa:
@c0d3z3r0, for that enumeration id (0xff00020f for "RAM1_01") can you copy and paste its contents from the enumeration tab, and also can you give me the output for the following:
```
import internal
print(internal.netnode.alt.repr(0xFF00020F))
```

It should look something like this on a 32-bit database:
```
print(internal.netnode.alt.repr(enum.by('ASN1encodingrule_e')))
fffffff8 : 0x124 (292)
fffffffb : 0x42 (66)
fffffffd : 0x1100000 (17825792)
ffffffff : 0x7 (7)
```

c0d3z3r0:
RAM1_01 enum:

~~~
[...]

PR #140, opened 2021-07-12
Reimplemented the `enumeration.has` function to avoid using altvals when determining whether an identifer belongs to an enumeration or not.

arizvisa:
The implementation of the `enumeration.has` function explicitly tests the netnode for the provided enumeration identifier in order to verify whether the given id is an enumeration or not. As reported in issue #139, these tests are too specific in that the "ENUM_ORDINAL" altval isn't always defined if it's the first enumeration. From reversing libida64.so, it was discovered that IDA actually uses other (undocumented) tag types as well a couple of other things which makes explicit testing of the netnode not a reliable way to determine an enumeration.

This PR adds support for specifying arbitrary tags to the all of the `internal.netnode` functions, and rewrites the `enumeration.has` function to remove the explicit messing around with netnodes from its implementation. Another case was added to the `enumeration.has` function so that enumerations can be checked by name.

This fixes issue #139.

### DIFF
--- a/base/_netnode.py
+++ b/base/_netnode.py
@@ -543,59 +543,59 @@ class sup(object):
     MAX_SIZE = 0x400
 
     @classmethod
-    def has(cls, nodeidx, index):
+    def has(cls, nodeidx, index, tag=None):
         '''Return whether the netnode identified by `nodeidx` has a "supval" for the specified `index`.'''
-        return any(index == item for item in cls.fiter(nodeidx))
+        return any(index == item for item in cls.fiter(nodeidx, tag=tag))
 
     @classmethod
-    def get(cls, nodeidx, index, type=None):
+    def get(cls, nodeidx, index, type=None, tag=None):
         '''Return the value at the `index` of the "supval" array belonging to the netnode identified by `nodeidx` casted as the specified `type`.'''
         node = netnode.get(nodeidx)
         if type in {None}:
-            return netnode.supval(node, index)
+            return netnode.supval(node, index, tag or netnode.suptag)
         elif issubclass(type, memoryview):
-            res = netnode.supval(node, index)
+            res = netnode.supval(node, index, tag or netnode.suptag)
             return res and memoryview(res)
         elif issubclass(type, bytes):
-            return netnode.supstr(node, index)
+            return netnode.supstr(node, index, tag or netnode.suptag)
         elif issubclass(type, six.string_types):
-            return netnode.supstr(node, index)
+            return netnode.supstr(node, index, tag or netnode.suptag)
         raise internal.exceptions.InvalidTypeOrValueError(u"{:s}.get({:#x}, {:#x}, type={!r}) : An unsupported type ({!r}) was requested for the netnode's supval.".format('.'.join([__name__, cls.__name__]), nodeidx, index, type, type))
 
     @classmethod
-    def set(cls, nodeidx, index, value):
+    def set(cls, nodeidx, index, value, tag=None):
         '''Assign the provided `value` to the specified `index` of the "supval" array belonging to the netnode identified by `nodeidx`.'''
         node = netnode.get(nodeidx)
-        return netnode.supset(node, index, value.tobytes() if isinstance(value, memoryview) else value)
+        return netnode.supset(node, index, value.tobytes() if isinstance(value, memoryview) else value, tag or netnode.suptag)
 
     @classmethod
-    def remove(cls, nodeidx, index):
+    def remove(cls, nodeidx, index, tag=None):
         '''Remove the value at the specified `index` of the "supval" array belonging to the netnode identified by `nodeidx`.'''
         node = netnode.get(nodeidx)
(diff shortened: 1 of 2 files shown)


base/_netnode.py
```diff
@@ -533,58 +539,58 @@ class sup(object):
...
         return
 
     @classmethod
-    def repr(cls, nodeidx):
+    def repr(cls, nodeidx, tag=None):
         '''Display the "supval" array belonging to the netnode identified by `nodeidx`.'''
         res = []
-        for index, item in enumerate(cls.fiter(nodeidx)):
+        for index, item in enumerate(cls.fiter(nodeidx, tag=tag)):
             value = cls.get(nodeidx, item)
```
arizvisa, line 594: This "cls.get" needs the "tag" parameter set when getting a "supval".

base/_netnode.py
```diff
@@ -474,49 +480,49 @@ class alt(object):
...
-        return netnode.altval(node, index)
+        return netnode.altval(node, index, tag or netnode.alttag)
 
     @classmethod
-    def set(cls, nodeidx, index, value):
+    def set(cls, nodeidx, index, value, tag=None):
         '''Assign the integer `value` at the `index` of the "altval" array belonging to the netnode identified by `nodeidx`.'''
         node = netnode.get(nodeidx)
-        return netnode.altset(node, index, value)
+        return netnode.altset(node, index, value or netnode.alttag)
```
arizvisa, line 497: The assignment of the "tag" here has been mistakenly merged with the "value".

base/_netnode.py
```diff
@@ -474,49 +480,49 @@ class alt(object):
...
-        return netnode.altdel(node, index)
+        return netnode.altdel(node, index, tag or netnode.alttag)
 
     @classmethod
-    def fiter(cls, nodeidx):
+    def fiter(cls, nodeidx, tag=None):
         '''Iterate through all of the elements of the "altval" array belonging to the netnode identified by `nodeidx` in order.'''
         node = netnode.get(nodeidx)
-        for index, value in utils.falt(node):
+        for index, value in utils.falt(node, tag=tag or netnode.alttag):
```
arizvisa, line 509: Pretty sure we can just assign the "tag" parameter and drop the conditional assignment for "netnode.alttag".

base/_netnode.py
```diff
@@ -474,49 +480,49 @@ class alt(object):
...
             yield index, value
         return
 
     @classmethod
-    def riter(cls, nodeidx):
+    def riter(cls, nodeidx, tag=None):
         '''Iterate through all of the elements of the "altval" array belonging to the netnode identified by `nodeidx` in reverse order.'''
         node = netnode.get(nodeidx)
-        for index, value in utils.ralt(node):
+        for index, value in utils.ralt(node, tag=tag or netnode.alttag):
```
arizvisa, line 517: We can assign the "tag" parameter and drop the conditional assignment for "netnode.alttag".

base/_netnode.py
```diff
@@ -533,58 +539,58 @@ class sup(object):
...
-        return netnode.supdel(node, index)
+        return netnode.supdel(node, index, tag or netnode.suptag)
 
     @classmethod
-    def fiter(cls, nodeidx):
+    def fiter(cls, nodeidx, tag=None):
         '''Iterate through all of the elements of the "supval" array belonging to the netnode identified by `nodeidx` in order.'''
         node = netnode.get(nodeidx)
-        for index, _ in utils.fsup(node):
+        for index, _ in utils.fsup(node, tag=tag or netnode.suptag):
```
arizvisa, line 577: I think we can drop the "netnode.suptag" conditional and just assign the "tag" parameter.

arizvisa (commented):
Some fixes that'll need to be force-push'd.

base/_netnode.py
arizvisa: not true.

base/_netnode.py
arizvisa: also not necessary.

base/_netnode.py
arizvisa: nope. not necessary.

base/_netnode.py
```diff
@@ -116,6 +116,12 @@ class netnode(object):
         altfirst = _ida_netnode.netnode_altfirst
         altnext = _ida_netnode.netnode_altnext
 
+    # default tags
+    alttag = idaapi.atag
+    suptag = idaapi.stag
+    hashtag = idaapi.htag
+    chartag = 0x64
```
arizvisa, line 123: I think on older versions of IDA, this can't be an integer and must be a char/byte.

arizvisa: Probably worth adding a comment describing where I reversed this value from.

arizvisa (commented):
One or two more things to fix...

arizvisa (commented):
This now looks good and ready to merge after confirmation.

arizvisa (commented):
Ready to merge.

Issue #141, opened 2021-08-22
typeinfo hooks raising exceptions


c0d3z3r0:
Getting these exceptions when using `ida-efitools2` with `ida-minsc` loaded. Commenting out that logging line, other ones in hooks.py fail, too. IDA Pro 7.6.210427

~~~
CRITICAL:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) : Callback for IDB_Hooks.changing_ti with priority (+0) raised an exception while executing <bound method type.changing of <class 'hooks.typeinfo'>>
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) : Traceback (IDB_Hooks.changing_ti was hooked at)
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) :   File "C:\Program Files\IDA7.5.20.1028\python\2\ida_idp.py", line 6066, in dispatch
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) :     cb.fun(slot, *args)
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) :   File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\ida_minsc\base\_interface.py", line 524, in closure
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) :     result = callable(*parameters)
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) :   File "C:\users\c0d3\Application Data\Hex-Rays\IDA Pro\ida_minsc\misc\hooks.py", line 1486, in make_ida_not_suck_cocks
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) :     ui.hook.idb.add('changing_ti', typeinfo.changing, 0)
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) : 
WARNING:root:internal.interface.priorityhook.callback(18374686481819107521L, '\n=\x04#\x8eu', None) : Traceback (most recent call last):
[...]

PR #142, opened 2021-09-01
Fixed all of the hooks involving type information so that the type information is properly escaped when logged.

arizvisa:
In order to cache any type information changes in the tag cache, the necessary events are hooked in order to ensure that the changes are kept track of. To help debugging with this, a number of logging events are created to describe what was changed and what is being logged. When formatting some of these log records, the type information (which is raw binary) is not escaped as a string thus resulting in an exception being raised while attempting to log the record.

Fixing this is simple as all we need to do is find the type information that is being rendered, and escape it with the `utils.string.repr` function to ensure it can be displayed in the string properly. This only happens in the "misc/hooks.py" module, and it only occurs in 8 places.

This fixes issue #141.

### DIFF
--- a/misc/hooks.py
+++ b/misc/hooks.py
@@ -639,21 +639,21 @@ class typeinfo(changingchanged):
     @classmethod
     def changing(cls, ea, new_type, new_fname):
         if not cls.is_ready():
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
Appears that the emitted type information is properly being escaped by `utils.string.repr`. Ready to merge.

arizvisa:
Merging.

Issue #143, opened 2021-10-06
Some of the 64-bit registers being used in operands are not being decoded properly.


arizvisa:
Turns out that as a remnant from when this plugin was 32-bit, I'm not actually decoding `idaapi.o_phrase` and possibly `idaapi.o_displ` operands correctly. Certain SIB operands that use `%r8` through `%r15` are being returned as their 32-bit register variations. This is due the 4th bit for the operands being stored in `insn_t.insnpref` variable.

Anyways, this results in operands like `(%rsi,%r14)` or `(%r14,%r15)` being decoded as `(%rsi,%rsi)` and `(%rsi, %rdi)` respectively. I'll need to refactor the parameter used for decoding so that it takes the `idaapi.insn_t` instead of just the address.

Meh... :-(

arizvisa:
This is being worked on in PR #144.

PR #144, opened 2021-10-07
Updated the instruction operand decoders so that they decode the phrase and memory types more accurately.

arizvisa:
When the instruction operand decoders were first written much earlier in this project's lifetime, there was a lot of information that just wasn't available at the time. As a result, a lot of the operand decoding methodology was inaccurate or missed some corner cases. As a result of issue #143, the operand decoders for the intel architecture were revisited in order to correct some of these known (and hopefully unknown) issues.

This resulted in modifying the hidden `__optype__` lookup class to take an instruction type instead of an address. Originally this had been done like this because it was considered by the author that another disassembler with a better documented API should be used, which ended up not occurring. All of the guards that were used by the operand decoders during their development have been culled as the implicit lookup that is done by `__optype__` should already guarantee the operand type is as expected. Address sizes no longer depend on the database size, instead taking their size from the "auxprefs" field of the `idaapi.insn_t`. The `intelops.SegmentOffset` class is now used by the operand decoder for the `idaapi.o_mem` type. This change might break backwards compatibility, but it's for the better.

The last change that was made was fixing support for 64-bit registers when decoding an `idaapi.o_displ` or `idaapi.o_phrase` type. The way this was being accomplished originally was by cross-checking the "dtype" field of the `idaapi.op_t` that was being decoded. It was determined that this was not sufficient due to the existence of issue #143. The decoder now will properly check the 4th bit that's hidden away in the "insnpref" field of the `idaapi.insn_t`, and use that to determine which 64-bit register should be returned. The checking of the "dtype" field has essentially been demoted to just a sanity check. There is still some more work that can be done with regards to the other architectures, but this should handle most of one's general decoding needs for now.

As a result of the decoding being fixed, the implementation of the `instruction.op_structure` function and a number of others will now use it to determine the offset for a given operand. Before this, the implementation was actually reading the operand attributes itself, and taking a guess for which property to use depending on the operand's type. It was decided to use the decoder directly in order to avoid any duplication of the need to clamp an operand value to its segment size. Another bug that resulted in the YMM registers being incorrectly sized was fixed, and the 64-bit CRx registers were included due to being missed when adding 64-bit register support.

Some aesthetic changes are also introduced via this PR. Prior to this, all instances of the `idaapi.insn_t` type were assigned to a variable "inst". For consistency (and hopefully clarity), these variables have now been renamed to "insn" so it directly corresponds to the type of their value. It was also discovered that the `interface.map_t` which is used to instantiate the `instruction.architecture.register` object was being incorrectly rendered to the wrong types being used in its formatspec. This was also corrected.

This fixes issue #143.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -1280,9 +1280,11 @@ class node(object):
         '''Return whether the operand `opnum` at the address `ea` has its sign inverted or not.'''
         AFL_SIGN0, AFL_SIGN1 = 0x100000, 0x200000
 
-        # Verify that we were given an operand number that has been tested, and
-        # log it if we haven't. Although it's likely that inverting the second
-        # operand will afect all of the following operands, we do this to be safe.
+        # Verify that we were given an operand number that has been tested before,
+        # and log it if we haven't. Although it's likely that IDA will consider
+        # all of the operands that follow the second operand as inverted once the
+        # inversion has been applied by the user, we log this just to be safe and
+        # let the user know that we're making an assumption here.
         if opnum not in {0, 1, 2}:
             result = cls.aflags(ea)
             logging.info(u"{:s}.alt_opinverted({:#x}, {:d}) : Fetching the inversion state for the operand ({:d}) of the instruction at {:#x} has not been tested (aflags={:#x}).".format('.'.join([__name__, cls.__name__]), ea, opnum, opnum, ea, result))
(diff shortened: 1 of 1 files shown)


arizvisa:
I guess I'ma merge this because I seriously forgot about it, but I've been using the branch for like a week and a half.

Issue #146, opened 2021-10-31
Minor issues with multicased functions that can result in matches not being calculated correctly.


arizvisa:
While dealing with a keyword parameter I encountered an issue that will manifest when you recurse into an alternative case incorrectly. This issue is just a placeholder for a reworking of some of the matching code so that we avoid depending on Python to calculate the placement of the various arguments.

arizvisa:
This issue is being fixed by PR #147, hence I'm linking this issue to that PR.

arizvisa:
This was merged.

PR #147, opened 2021-10-31
Fixed the way that the multicase decorator works so that all the arguments are properly checked against their constraints when matching.

arizvisa:
While working on some of the navigators inside the `database.address` namespace, it was discovered that some of the arguments types were being checked more than once. More importantly, however, the arguments types were being collected prior to returning them to the caller.. yet, in all actuality, they were not being used when dispatching to the matched function. Instead what was happening, was that the original arguments were then being passed to the matched function using python's internal function argument dispatcher and the original parameters the user gave us. Although this still worked properly, it was not actually how I originally intended the decorator to be used.

Due to the way a particular condtional was being used in `utils.multicase.match`, any functions that were matched with duplicate parameter names would not have their first parameter (the class or instance) checked. This was due to the function being incorrectly decorated. Again though, this didn't affect anything because the first parameter was being discarded and ignored if the function was decorated as an instancemethod or a classmethod. Nonetheless, this was fixed so that if the user wants they can also specify a type constraint for the class or instance that's being passed as the first parameter instead of us just ignoring it like we were doing previously.

The decorated documentation that is generated was also fixed to remove the first parameter that's implicitly included. As this parameter is generally irrelevant, the displaying of it along with its type takes up extra unnecessary space which would be better if it was occupied by actual relevant information.

This fixes issue #146.

### DIFF
--- a/base/_utils.py
+++ b/base/_utils.py
@@ -545,13 +545,13 @@ class multicase(object):
             # decorated with. If our constraint is a builtins.callable, then we just need to
             # ensure that the parameter can be called. Otherwise our constraint should be an
             # iterable of types that we can simply pass long to the isinstance() function.
-            critiqueF = lambda constraint: builtins.callable if constraint == builtins.callable else finstance(*constraint)
+            critiqueF = lambda constraint: builtins.callable if constraint == builtins.callable else frpartial(builtins.isinstance, constraint)
 
             # Zip our parameter names along with our argument values so that we can extract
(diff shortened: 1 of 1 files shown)


arizvisa:
I think that's it, just need to review this fucker.

arizvisa (commented):
This PR was too much of a tangent. It's been reviewed, albeit lazily. So yeah, this fucking looks okay and I'm merging it because I don't care right now.

Issue #149, opened 2021-12-14
Internal object representing the library for the native lower-level api does not support the 64-bit OSX library


arizvisa:
As reported by @binerdd in private, the meta_path finder object which is used to to create a ctypes library to the native API does not support OSX. This is due to an oversight by the author where they neglected to include the 64-bit filename that is used by OSX and thus only the 32-bit filename is used to create the ctypes library.

This is a simple fix, namely the filenames used by IDA on OSX is:

`libida.dylib` for 32-bit, and `libida64.dylib` for 64-bit.

arizvisa:
This is fixed by PR #150.

arizvisa:
This was fixed by PR #150 which was just merged. Closing this issue.

PR #150, opened 2021-12-14
Fixed the `meta_path` used for the darwin platform to add the 64-bit variation of the shared object filename

arizvisa:
When the plugin initializes its base namespace, it uses Python's `meta_path` to add a finder that allows the user to interact with the native IDA shared object using the `ctypes` library. This allows one to interact with the exports that IDA's API exposes to a user in the situation that the desired functionality is not exposed by the IDAPython SWIG interface.

It was identified that the library on the "darwin" platfrom was incomplete and only included the path for the 32-bit version of IDA. This was due to an oversight by the author and was fixed by testing whether the 32-bit or 64-bit version of IDA is running and choosing the correct filename based on the detected version.

On the "darwin" platform, the 32-bit version of the library is "libida.dylib" and the 64-bit version of the library is "libida64.dylib". Similar to the way the internal is determined for the other platforms, the value for `idaapi.BADADDR` is checked and used to determine whether to add the "64" suffix or not. The "linux" platform was also updated to use `idaapi.idadir` to determine the installation directory and locate the correct native library.

This fixes issue #149.

### DIFF
--- a/idapythonrc.py
+++ b/idapythonrc.py
@@ -230,7 +230,7 @@ if sys.platform in {'darwin'}:
     sys.meta_path.append( internal_object('ida', library(idaapi.idadir("libida{:s}.dylib".format('' if idaapi.BADADDR < 0x100000000 else '64')))) )
 
 elif sys.platform in {'linux', 'linux2'}:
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
This looks good. It was only minor, anyways.

Issue #152, opened 2022-02-08
Uninitialized `CRITICAL_SECTION` dereferenced by `_ida_idp.pyd` from IDA 7.7 when interacting with a database or using certain hooks from `IDP_Hooks` and `UI_Hooks`.


shavitush:
Using IDA Pro 7.7 SP1 and Python 3.9.10 on Windows. Happens on both x86 and x86-64 databases.

Used `idapyswitch` in IDA's root folder to use the version of Python I installed myself rather than the one bundled with IDA. I've installed using the instructions in README - cloned the repository into `%user%\AppData\Roaming\Hex-Rays\IDA Pro`, ran `pip install -r 'requirements.txt'` and it's loaded. However IDA's output window says;

> WARNING:root:No .idapythonrc.py file found in the user's home directory (C:\Users\my_name).

(which isn't even the full path to my home directory according to `print(idaapi.get_user_idadir())`)

The output window also greets me for using ida-minsc. Invoking `database.config.version()` as stated in the README returns an unexpected 0x302 when a database is not loaded, rather the expected 0 even though 770 would make sense when one is loaded.

![image](https://user-images.githubusercontent.com/3672466/152947732-92d96dd9-0511-4198-beb6-862b41b5ca27.png)

Then IDA crashes as soon as any database is loaded.

![image](https://user-images.githubusercontent.com/3672466/152948060-2e28fc75-6466-4d92-9dae-b5972c79c104.png)

arizvisa:
Um.... Wow, it's straight-up crashing...during database-load? I think that's happening when choosing the processor module?? Is that really the same for all binaries?

arizvisa:
When you startup ida, try typing in `ui.hook.idp` that should list all of the hooks. You're going to be disabling the processor hook. That's done with:
```
ui.hook.idp.disable('ev_newprc')
```

At that point, when you open up your database, the processor hook won't run. All that processor hook is doing is figuring out which instruction set and registers you're using. That's being done by checking the `idaapi.ph.id` attribute as in various versions of IDA the id that's passed to newprc can be wrong.

Anyways, after the database loads. Try running the following which is actually how minsc expects IDA to call the hook.
```
processor_id = idaapi.ph.id
ins.__newprc__(processor_id, True)
```

Hopefully it crashes during one of these things.

shavitush:
Just tried that, it lists:
> IDP_Hooks.ev_newprc (disabled)

[...]

PR #153, opened 2022-02-09
Refactored the `prioritybase` related classes and specifically `priorityhook` so that hooking is done on-demand to prevent a null-pointer dereference.

arizvisa:
On some instances of IDA 7.7 on Windows, a null-pointer dereference may occur if certain hooks are used. Technically the plugin actually implements all of the available hooks within the `IDP_Hooks`, `IDB_Hooks`, and `UI_Hooks` classes in order to allow a user to track down which events are dispatched while executing a particular action with the idea that they could capture events, hook, and replay them.

Unfortunately due to the issue reported in #152, it is exactly this functionality that's resulting in an uninitialized `CRITICAL_SECTION` being used. In order to quickly (or not-so-quickly) come up with a fix, these problematic events were enumerated through a combination of debug logging and subdivision of the available hooks to narrow down on what was being called. With this I was able to put together a blacklist of the malfunctioning event hooks which resulted in commit b309c4cb74fc52bf9b9ab31c4c20a1a96deb9fa4.

At first it was thought that this was to be some re-entrancy issue or something because it's hooking, and crashing due to a NULL being passed to `EnterCriticalSection`. However, after debugging it a few minutes ago. It turns out that it's a straight-up dereference of a `CRITICAL_SECTION` that hasn't yet been initialized. The logic for this is performed by the `_ida_idp.pyd` python module. Setting a breakpoint on module load showed that it's never even written to. Pulling it into the disassembler, it's a global that turns out to contain no other direct references. Looking before it for any other boundaries to ensure that it's not an object/structure or similar turned up no results. The caller of the function that dereferences the `CRITICAL_SECTION` in question is referenced 10 times which likely corresponds to 10 events that do not work anymore. Only 8 events were enumerated, however, and were listed in the issue at #152.

This PR fixes the related issue by refactoring the way the hooks work so that they're \_always\_ done on-demand. This involved some refactoring in both `base/_interface.py` and `misc/ui.py` and resulted in the hiding of the `prioritybase.hook` and `prioritybase.unhook` methods in order to simplify the interface that the user has access to. Instead we handle the scoping of these hooks on behalf of the user so that they don't have to activate or deactivate anything, and we just try to do proper cleanup on IDA's exit. Some properties were also added so that users can enumerate all of the available hooks if they want to capture events as they're happening, and some warnings will probably be added later to let the user know that these 10 particular hooks cannot be used in 7.7.

This fixes issue #152.


arizvisa:
This breaks when using Python2 due to the way a closure is capturing variables inside `base/_interface.py`. It also appears that the `on_init` hook which kickstarts everything is not working reliably on 7.1.

arizvisa:
force-push'd, haven't done the review yet. but gonna ping the reporter first.

arizvisa:
Another force push.. This time it's due to a rebase on top of commits 6df77492314cea2a80d9c53ab4e13c28b6a266c2 and 80baf902332dbaf5478a1b0ee3e2a623ddc610f6.

arizvisa:
Reporter of issue confirmed this PR fixes the issue. Just need to do a review.

arizvisa (commented):
Some public-only APIs are being used for internal usage.

base/_interface.py
```diff
@@ -635,233 +628,205 @@ class priorityhook(prioritybase):
...
+            attach.update(self.__attached__)
+        return super(priorityhook, self).detach(name)
 
-    def add(self, target, callable, priority):
-        '''Add the `callable` to the queue for the specified `target` with the provided `priority`.'''
-        if target in self.available:
-            return super(priorityhook, self).add(target, callable, priority)
+    def add(self, name, callable, priority):
+        '''Add the `callable` to the queue for the specified `name` with the provided `priority`.'''
+        if name in self.available:
```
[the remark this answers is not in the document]
arizvisa: done.

base/_interface.py
```diff
@@ -302,73 +302,95 @@ def __init__(self):
         self.__traceback = {}
 
     def __iter__(self):
-        '''Return the id of each target that is hooked by this object.'''
+        '''Return the each target that is attached to this object.'''
```
arizvisa, line 305: speak english, dude.

base/_interface.py
```diff
@@ -302,73 +302,95 @@ def __init__(self):
...
         for target in self.__cache__:
             yield target
         return
 
+    def __contains__(self, target):
+        '''Return whether the provied `target` is currently attached to this object.'''
```
arizvisa, line 311: spelling.

arizvisa (commented):
meh.

arizvisa:
I am just merging this. I don't care anymore.

Issue #154, opened 2022-02-11
Anything that accesses `func_t.referers` from `idaapi.get_fchunk` when `func_t.flags & idaapi.FUNC_TAIL` is set can result in a null pointer dereference


arizvisa:
In order to check a function tail, one is supposed to use `idaapi.get_fchunk` to get a `func_t`. Once getting an instance of this type, you're then supposed to check the flags (`func_t.flags`) for the `idaapi.FUNC_TAIL` bit. At this point the `func_t.referers` attribute contains a `dyn_ea_array` of all of the referrers of this particular chunk. You can check the number of referrers by accessing `func_t.referers.count`.

In some cases when the count is larger than 0 (it should always be larger than 0), accessing the first element can result in a null-pointer dereference. This occurs in `_ida_funcs.pyd`. There's a number of places that a tail's referrers are actually checked. All of these places will likely cause IDAPython to crash.

arizvisa:
I'm using the following logic in this database DBCLIENT.DLL (6.84.8027.1) from some random software to reproduce:

```
a = [0x1800a45dc, 0x1800a4611, 0x1800a4659, 0x1800a468b, 0x1800a5be8, 0x1800a5c17, 0x1800a5c4b]
for ea, f in zip(a, map(idaapi.get_fchunk, a)):
    if f and f.flags & idaapi.FUNC_TAIL and f.referers.count == 1:
        print(f.referers[0])
    continue
```

arizvisa:
This is fixed by PR #155.

arizvisa:
Reflecting on the fix I wrote for this, I probably should be using `idaapi.func_parent_iterator_t`... :-/

Nonetheless, closing this PR as #155 was merged.

PR #155, opened 2022-02-11
Modified all instances where the "referers" of a function tail is checked to verify the quantity in "refqty" before accessing the property.

arizvisa:
It seems that the `func_t.referers` attribute is super unstable and almost racy when it is fetched by `idaapi.get_fchunk`. What you're supposed to do is use `idaapi.get_fchunk` to get a `func_t`, then check the `func_t.flags` for the `FUNC_TAIL` flag. If this is set, then the `func_t.referers` attribute contains a list of the referrers (or owners) for that particular function chunk. This attribute has a "count" attribute which you can use to check how many elements are in the array. There are instances where the "count" attribute is set to 1, but as soon as you access the first element of the array it dereferences a null-pointer. Historically, this attribute has been super-unstable but you could work around it in most cases. It seems that now with 7.7 it is as unreliable as all hell.

This PR works around this issue by checking the `func_t.refqty` attribute before accessing the `func_t.referers` attribute. If the `func_t.refqty` attribute is 1, then it completely avoids accessing the referers instead and falling back to `idaapi.get_func` to get the owner. This will work for most situations, but anytime the number of referers is greater than 1, this risks the null pointer dereference. I'm reporting the issue to support@hex-rays.com, but instead of waiting will probably push this fix as I still need to do my regular job...

This fixes issue #154.

### DIFF
--- a/base/_comment.py
+++ b/base/_comment.py
@@ -831,15 +831,15 @@ class contents(tagging):
                 logging.critical(u"{:s}._key({:#x}) : Unable to read more than one referrer ({:d}) for the function tail at {!s}. The \"{:s}\" attribute is {!s}.".format('.'.join([__name__, cls.__name__]), ea, count, bounds, 'referers', ch.referers))
                 return internal.interface.range.start(res)
 
-            # Otherwise we can just use idaapi.get_func() to get the owner.
-            elif ch.referers is None:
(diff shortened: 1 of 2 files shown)


arizvisa (commented):
Small fix. Easy to review.

Issue #156, opened 2022-05-23
The pythonic `type` doesn't seem to work when adding or assigning it to a field 


arizvisa:
As reported by someone in private, the pythonic `type` does not actually work when adding it to a field. This is actually because IDA deals with this specially (using `FF_`-based flags which aren't based on the size). So, when I implemented it...it was definitely hacked into place.

This involves fixing two things, one of them is the typemap dissolver and the other is related to why IDA throws up a messagebox whenever I try to assign flags to the type.

arizvisa:
The mapping of the `type` keyword seems to have been fixed, but the API that's being used to fetch the type seems to be raising a warning box after the type has been applied. The warning that's being displayed is "struct->til conversion failed". I imagine this has to do with the flags that're being applied as in my tests only the MS_1TYPE flag seems to apply to a field and I'm pretty sure I'm setting both to deal with things other than structure members. Hence, this isn't entirely fixed and will still need looking into.

For now, PR #157 fixes the mapping of the type.

arizvisa:
Closing as this was fixed by PR #157.

arizvisa:
Apparently the `database.get.structure` function was not decoding the pythonic pointer type correctly. This was fixed by commit 84112203c2cde2e48d1913417a8602b0e70a3afd.

PR #157, opened 2022-05-23
Fixed the mapping of `type` to the pointer type which is used for describing structure fields.

arizvisa:
The mapping of the `type` field used by both the `structure` module and some of the functions within the `database` module was not correctly being mapped to the flags that are applied to a field. This seems to manifest itself in two ways. One of which is that the size calculation for the boundaries of a type is not correct. The other way is that once the type for a field has been changed, the refinfo for the field needs to be updated.

The size calculation issue was resolved by fixing the `typemap.dissolve` implementation in the `internal.interface` module and its tables. As the offset type is not actually a size (`DT_TYPE`) and it includes operand information, the implementation needed to also include `MS_0TYPE` and `MS_1TYPE` in its checks. This was done by adding two variables, `dsize` (which contained the original value from `DT_TYPE`), and `dtype` which includes the operand information from the flags. As a result, the `inverted` table were also updated to add all 3 operand variations of the flags for each of the sizes so that they can be looked up to determine the type. Another issue that was resolved was that apparently IDA does not use `BADADDR` inside the `opinfo_t` when pointers are being used which was resolved by setting `opinfo_t.tid` to 0 when a pointer type gets created.

The reference information was resolved by the introduction of the `typemap.update_refinfo` function which calculates which operands for a given address are set, and then updates the address' reference information using the correct size as determined from the flags (`DT_TYPE`).

After these additions, the `database.set.array`function , the `structure.members_t.add` method, and the `structure.member_t.type` property all needed to be updated so that the reference gets added. Unfortunately, the view of the database is not being updated which makes it appear as if the type wasn't applied despite it being correct in the database. This addition also results in any array within the database being treated as an offset by default. This is because IDA sets the reftype for all arrays anyways if the integer actually references something in the database. For the record, there's no explicit support for user-defined reference types without using the `refinfo_desc_t` as a map, so using the size to figure this out is literally the only viable option.

Fixes issue #156.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -74,10 +74,6 @@ class typemap(object):
     FF_MASKSIZE = idaapi.as_uint32(idaapi.DT_TYPE)  # Mask that select's the flag's size
     FF_MASK = FF_MASKSIZE | MS_0TYPE | MS_1TYPE     # Mask that select's the flag's repr
 
-    # FIXME: In some cases FF_nOFF (where n is 0 or 1) does not actually
-    #        get auto-treated as an pointer by ida. Instead, it appears to
-    #        only get marked as an "offset" and rendered as an integer.
-
     # FIXME: Figure out how to update this to use/create an idaapi.tinfo_t()
     #        and also still remain backwards-compatible with the older idaapi.opinfo_t()
 
@@ -168,6 +164,12 @@ class typemap(object):
     # FIXME: this is a hack for dealing with structures that
     #        have the flag set but aren't actually structures..
     inverted[idaapi.FF_STRUCT if hasattr(idaapi, 'FF_STRUCT') else idaapi.FF_STRU] = (int, 1)
+
+    # refinfo map for the sizes (IDA 6.9 uses the same names)
+    refinfomap = {
+        1 : idaapi.REF_OFF8,    2 : idaapi.REF_OFF16,
+        4 : idaapi.REF_OFF32,   8 : idaapi.REF_OFF64,
+    }
 
     # Assign the default values for the processor that was selected for the database.
     @classmethod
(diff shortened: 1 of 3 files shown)


arizvisa:
So, applying the flags for the pointer type has now been corrected, and I also enabled support for inverting the flags for alignment to the `None` type. This should correct the semantics that are described by the documentation in both the `database` and the `structure` modules.

However, my theory on setting the flags incorrectly was wrong, and in all actuality it's because IDA wants `opinfo_t.tid` to be set to 0 when setting either of the `FF_0OFF` or `FF_1OFF` flags on a member. Although this "resolves" and "dissolves" the types correctly, it seems that once the type has been correctly changed that the currently applied structure does not try to resolve the field to its actual target.

From what I can tell, the only difference is literally NSUP_REF0 is set to "\x10\x00" when adding the field with `idaapi.add_struc_member`. This is different than explicitly setting it with "o" to mark it as an offset because that results in NSUP_REF0 being set to "\x02". This is related to the refinfo stuff inside IDA, but it is unfortunately not reachable from the pythonic-type implementation inside the `typemap` namespace. I need to do some more research here, but I'm pretty sure I'll need to tweak anywhere that pythonic-types get used so that they also update NSUP_REF0 with whatever api I'm "supposed" to use.

arizvisa:
Okay, so here's what's missing that needs to be set to a field in order to make it an actual pointer. NSUP_REFX actually refers to the refinfo_t. The value that I was looking at appears to be the reftype_t which has one of the following values from nalt.hpp:
```
  V695_REF_OFF8 = 0,      ///< reserved
  REF_OFF16  = 1,         ///< 16bit full offset
  REF_OFF32  = 2,         ///< 32bit full offset
  REF_LOW8   = 3,         ///< low 8bits of 16bit offset
  REF_LOW16  = 4,         ///< low 16bits of 32bit offset
  REF_HIGH8  = 5,         ///< high 8bits of 16bit offset
  REF_HIGH16 = 6,         ///< high 16bits of 32bit offset
  V695_REF_VHIGH  = 7,    ///< obsolete
  V695_REF_VLOW   = 8,    ///< obsolete
  REF_OFF64  = 9,         ///< 64bit full offset
  REF_OFF8   = 10,        ///< 8bit full offset
  REF_LAST = REF_OFF8;
```
I can set this via `idaapi.set_refinfo(ea, opnum, type, target=BADADDR, base=0, tdelta=0)` which should probably be done to the entire pythonic-types api so that you can assign them properly based on the size. This means, `REF_OFF8`, `REF_OFF16`, `REF_OFF32`, and `REF_OFF64` will need to be put into use. One issue, though, with setting the refinfo is that the database (the line item, specifically) needs to be redrawn for it to immediately change once it's been applied.

Instead of making it part of pythonic-types, I'm likely considering just explicitly setting the refinfo using the size that `typemap.resolve` returns for whatever it is that the user has decided to apply their type to.

arizvisa:
Added the reference information functionality as `internal.interface.typemap.refinfo`, and updated both the `structure` and `database` modules to actually use it.

base/_interface.py
```diff
@@ -189,32 +197,33 @@ def dissolve(cls, flag, typeid, size):
...
-        dt = flag & cls.FF_MASKSIZE
+        dtype, dsize = flag & cls.FF_MASK, flag & cls.FF_MASKSIZE
         sf = -1 if flag & idaapi.FF_SIGN == idaapi.FF_SIGN else +1
 
         # Check if the dtype is a structure and our type-id is an integer so that we
         # figure out the structure's size. We also do an explicit check if the type-id
         # is a structure because in some cases, IDA will forget to set the FF_STRUCT
         # flag but still assign the structure type-id to a union member.
-        if (dt == FF_STRUCT and isinstance(typeid, six.integer_types)) or (typeid is not None and structure.has(typeid)):
+        if (dtype == FF_STRUCT and isinstance(typeid, six.integer_types)) or (typeid is not None and structure.has(typeid)):
```
arizvisa, line 207: This is wrong, and should be checking dsize (0xF0000000) instead of dtype (0xFFF00000).

base/_interface.py
```diff
@@ -189,32 +197,33 @@ def dissolve(cls, flag, typeid, size):
...
+        # type's size from the inverted map that we previously created. We start
+        # by checking the dtype first for pointers and then fall back to the size.
+        t, sz = cls.inverted[dtype] if dtype in cls.inverted else cls.inverted[dsize]
 
         # If the datatype size is not an integer, then we need to calculate the
         # size ourselves using the size parameter we were given and the element
         # size of the datatype that we extracted from the flags.
         if not isinstance(sz, six.integer_types):
-            count = size // idaapi.get_data_elsize(idaapi.BADADDR, dt, idaapi.opinfo_t())
+            count = size // idaapi.get_data_elsize(idaapi.BADADDR, dtype, idaapi.opinfo_t())
```
arizvisa, line 226: This also should be checking dsize (0xF0000000) instead of dtype (0xFFF00000).

arizvisa (commented):
Variable naming has resulted in a change of logic that should be corrected.

Issue #159, opened 2023-03-03
Miscellaneous compatibility issues with regards to IDA 6.8


gool123456:
Hi , My idol  💯 

I used IDA6.8 to download the main branches and reply to some error messages, including :
```

CRITICAL:root:internal : Error trying to import module `interface` from C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_interface.py.
Traceback (most recent call last):
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\plugins\minsc.py", line 159, in load_module
    res = self.new_api(name, path)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\plugins\minsc.py", line 133, in new_api
    res = super(cls, self).new_api(fullname, path)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\plugins\minsc.py", line 64, in new_api
    return self.imp.load_module(modulename, file, path, description)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_interface.py", line 18, in <module>
    class typemap(object):
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_interface.py", line 132, in typemap
    ptrmap = { (type, sz) : (idaapi.offflag() | flg, 0) for (_, sz), (flg, _) in integermap.items() }
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\base\_interface.py", line 132, in <dictcomp>
    ptrmap = { (type, sz) : (idaapi.offflag() | flg, 0) for (_, sz), (flg, _) in integermap.items() }
TypeError: 'int' object is not iterable
cannot import name interface
Traceback (most recent call last):
  File "D:\tools\IDA_Pro_v6.8_and_Hex-Rays_Decompiler_(ARM,x64,x86)_Hskyre\python\idaapi.py", line 641, in IDAPython_ExecScript
    execfile(script, g)
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\idapythonrc.py", line 45, in <module>
    loader.load(globals(), preserve={'loader', '_orig_stdout', '_orig_stderr'})
  File "C:\Users\Hskyre\AppData\Roaming\Hex-Rays\IDA Pro\plugins\minsc.py", line 254, in load
[...]

PR #160, opened 2023-03-05
Fixed a number of compatibility issues with older versions of the disassembler.

arizvisa:
As per issue #159, there were a number of random issues during load that occurred due to me not testing the codebase on older versions of the disassembler. Some of the issues involved misspellings, type-o's, and busted api usage. This PR fixes a number of these things, but stil leaves a couple of things broken in older versions. This is because there's is ongoing development going on in another branch and at the moment that branch gets priority.

Some of the things that are fixed by the PR are the type mapper within the `internal.interface.typemap` module, the hook which is used to track when the database is being closed, and usage of the wrong api within the `ui.current.selection` function. The main work within this PR actually involves the `internal.netnode` module with regards to the `_ida_netnode.exists` api not existing within IDAPython. Another issue this module had was that the closure that was being constructed for `netnode.exist_name` was not being decorated as a static-method which made it impossible to call correctly. Both of these functions were fixed by first decorating them as static methods, and then using `ctypes` to call into the SDK library to get access to the correct result.

This PR is actually somewhat related to commits 508a89ff203a9a6707645b624eb454bc8ecd470e and 10030d724a18795d02cda0e0c323ce9be29b4778. These commits aren't included in this PR because they weren't mentioned in the original report and were discovered due to having to forward-port the commits from this PR into the [persistence-refactor](https://github.com/arizvisa/ida-minsc/commits/persistence-refactor) branch.

Fixes issue #159.

### DIFF
--- a/base/_interface.py
+++ b/base/_interface.py
@@ -89,7 +89,7 @@ class typemap(object):
             (int, 16):(idaapi.owrdflag(), -1),
         }
         if hasattr(idaapi, 'ywrdflag'):
-            integermap[32] = getattr(idaapi, 'ywrdflag')(), -1
+            integermap[int, 32] = getattr(idaapi, 'ywrdflag')(), -1
 
         decimalmap = {
             (float,  4):(idaapi.floatflag(), -1),     (float, 8):(idaapi.doubleflag(), -1),
@@ -109,14 +109,14 @@ class typemap(object):
             (chr, 2): (idaapi.asciflag(), idaapi.ASCSTR_UNICODE),
 
             # variable-terminated, multiple-byte
-            (str, 1, 0): (idaapi.asciflag(), idaapi.ASCSTR_C),
+            (str, 1, 0): (idaapi.asciflag(), idaapi.ASCSTR_TERMCHR),
             (str, 2, 0): (idaapi.asciflag(), idaapi.ASCSTR_UNICODE),
             (str, 1, 1): (idaapi.asciflag(), idaapi.ASCSTR_PASCAL),
             (str, 1, 2): (idaapi.asciflag(), idaapi.ASCSTR_LEN2),
             (str, 2, 2): (idaapi.asciflag(), idaapi.ASCSTR_ULEN2),
             (str, 2, 4): (idaapi.asciflag(), idaapi.ASCSTR_ULEN4),
 
-            (chr, 1, 0): (idaapi.asciflag(), idaapi.ASCSTR_C),
+            (chr, 1, 0): (idaapi.asciflag(), idaapi.ASCSTR_TERMCHR),
             (chr, 2, 0): (idaapi.asciflag(), idaapi.ASCSTR_UNICODE),
             (chr, 1, 1): (idaapi.asciflag(), idaapi.ASCSTR_PASCAL),
             (chr, 1, 2): (idaapi.asciflag(), idaapi.ASCSTR_LEN2),
--- a/base/_netnode.py
+++ b/base/_netnode.py
@@ -36,17 +36,27 @@ class netnode(object):
     index = _ida_netnode.netnode_index
     kill = _ida_netnode.netnode_kill
 
-    # Although the following apis exist in 5.6 of the sdk, there's a chance that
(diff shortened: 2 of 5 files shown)


arizvisa:
Okay, so this PR fixed a number of compatibility issues that were reported by issue #159. This PR fixes the following issues:

1. The `internal.interface.typemap` namespace had suffered from copypasta when assigning the keys for the string types in older  versions of the disassembler. This resulted in the `ywrdflag` type being inaccessible and raising an exception when it occurs and exists. The string types were using `idaapi.ASCSTR_C` instead of the correct `idaapi.ASCSTR_TERMCHR`.
2. The `internal.netnode` module had issues due to the unavailability of the `idaapi.exist` api via IDAPython and the closures within the `netnode` namespace not being decorated as `staticmethod`. Both the `netnode.exist` and `netnode.exist_name` functions were refactored to always use an identifier, and `netnode.exist` was rewritten to fall back to using the `ida.netnode_exist` sdk function using "ctypes".
3. Some of the hooks used by older versions of the disassembler were using the wrong variable names or the wrong hook. All of logging done by the function-related hooks were tweaked to include the full boundaries of the function that they were being called with. This guarantees that the user can see the function being processed if an exception gets raised.
4. The `hooks.on_close` hook was being applied to `IDB_Hooks.closebase`. This is correct on later versions of the disassembler, but on older versions it should be `IDP_Hooks.closebase`. This was patched by determining which hook type has "closebase" available, and then using that one.
5. The `add_func` hook was also modified to check just the segment type when determining whether a `func_t` is added to an import or not instead of checking both. I distinctly remember encountering a situation where a `func_t` was applied to an import address that was inside a segment that wasn't of the `SEG_XTRN` type, but this doesn't actually make sense. More importantly, I don't know of an efficient way to detect import functions created in a non-external segment. If this issue does manifest again, it might only be fixable in the persistence branch where it's significantly easier to track database-scoped information.
6. Older versions of the disassembler have different semantics for the `idaapi.read_selection` function which affected the `ui.current.selection` function. This issue was discovered because older versions do not implement the `place_t.toea` method which is used to convert a `place_t` to an address. Instead, the `idaapi.read_selection` function returns a tuple composed of 3 values where the first value is a true or false that represents whether the api call was successful. Fixing this required a version guard that uses the `place_t` to determine the selected address range using the `place_t.toea` method in newer versions of the disassembler. In older versions, we assign everything that is returned by `idaapi.read_selection` and only use the addresses if the first element of the returned tuple is true.
7. The final issue which took some work to troubleshoot was a codepage issue (Thanks to @gool123456 for the help with reporting it and troubleshooting it). The plugin uses two functions within the `internal.utils.string` namespace to convert bytes returned from IDAPython to strings (and vice-versa). These two functions assume that all strings are using UTF-8 encoding as since I'm on Linux, that's what it uses. However, earlier versions of IDA on Windows do not use UTF-8 for its strings at all, instead opting to use the `CP_OEM` codepage according to documentation (and the default `CP_ACP` in all of my tests). This should work normally (since UTF-8 is backwards-compatible with ASCII), however, this isn't the case when users explicitly patch their disassembler for their alternate codepage. Fortunately, the disassembler provides a `get_codepages` function that one can use to determine which codepage the database is using. As this function is not available via IDAPython, it needs to be reached using "ctypes" with the SDK library. Since the codepage returned by Microsoft's API doesn't directly correspond to a codec type that can be looked up with Python, this resulted in the creation of the `string.codepage` namespace. This namespace contains a mapping that is used to correlate the Microsoft codepage (returned by the api) with the Python codec that will be used for encoding and decoding strings. The codec is then used to return one closure for decoding bytes to a string and another for encoding a string into bytes. This detection logic is only needed on the Windows platform and for versions of IDA prior to their i18n fixes that was documented at https://hex-rays.com/products/ida/news/7_0/docs/i18n/.

The additional usage of "ctypes" for this PR has convinced me that the upcoming work in "persistence-branch" should include a separate module that is responsible for assigning all of the necessary return and argument types for the functions within the `ida` module.

arizvisa (commented):
guess it's reviewed...

Issue #167, opened 2023-12-06
Suggestion: Add the name DLL as that is what IDA calls the shared objects


Harding-Stardust:
As a windows RE I am missing the name "is_dll()" as that is what IDA calls shared objects: 
```
ida_idaapi.get_inf_structure().is_dll()
```
https://github.com/arizvisa/ida-minsc/blob/af07f167a0992aed3f3a2a1670593880baa57820/base/database.py#L187

arizvisa:
I can totally add that alias for you. It makes sense. (Ftr, I'm a windows RE also).

arizvisa:
This was closed by PR #186.

PR #186, opened 2023-12-07
Added another alias to the `config.is_sharedobject` function as `config.is_dll` for the people whom are windows-centric.

arizvisa:
As requested in issue #167, the `config.is_sharedobject` naming might be unfamiliar for users that may have originated exclusively from the windows platform. This PR simply adds an alias for it as `config.is_dll`. Generally the contents of the namespace that this function resides within is unused, but the disassembler uses the naming scheme that is being discussed. Hence, it makes sense to add this alias as a just-in-case mechanism.

This fixes issue #167.


arizvisa (commented):
This is just a one-liner.

Issue #175, opened 2023-12-06
database.functions.list() throws exception, the user probably wanted to list all functions


Harding-Stardust:
database.functions.list() throws exception, the user probably wanted to list all functions.
```
database.functions.list()
internal.interface.tinfo.function_details(0x401029, '') : The type that was received ('') for the specified function (0x401029) was not a function type.
```
Also database.functions.list(like='*') throws Exception:
```
database.functions.list(like='*')
InvalidTypeOrValueError: internal.interface.tinfo.function_details(0x401029, '') : The type that was received ('') for the specified function (0x401029) was not a function type.
```

https://github.com/arizvisa/ida-minsc/blob/af07f167a0992aed3f3a2a1670593880baa57820/base/database.py#L595C11-L595C11

Harding-Stardust:
I get the correct tinfo_t with the following code:
```
type_at_addr = ida_typeinf.print_type(0x401029, ida_typeinf.PRTYPE_1LINE | ida_typeinf.PRTYPE_SEMI)
print(type_at_addr)

INT_PTR __userpurge DialogFunc@<eax>(INT_PTR@<eax>, HWND hDlg, UINT, WPARAM, LPARAM);
```

arizvisa:
Hmm. Trying to build a fresh database from the binary results in the following type being applied by default. For some reason, that also returns a valid type using your unit-test. :-/

```
.text:00401029
.text:00401029                                   ; Attributes: bp-based frame
.text:00401029
.text:00401029                                   ; INT_PTR __stdcall DialogFunc(HWND, UINT, WPARAM, LPARAM)
.text:00401029                                   DialogFunc      proc near               ; DATA XREF: start+E↑o
.text:00401029
.text:00401029                                   hDlg            = dword ptr  8
.text:00401029                                   arg_4           = dword ptr  0Ch
.text:00401029                                   arg_8           = dword ptr  10h

Python>function.type(0x401029)
INT_PTR __stdcall(HWND, UINT, WPARAM, LPARAM)
```

Applying the `__userpurge` type directly and trying again still returns the same....
```
.text:00401029                                   ; Attributes: bp-based frame
.text:00401029
.text:00401029                                   ; INT_PTR __userpurge DialogFunc@<eax>(INT_PTR@<eax>, HWND hDlg, UINT, WPARAM, LPARAM)
.text:00401029                                   DialogFunc      proc near               ; DATA XREF: start+E↑o
.text:00401029
[...]

PR #187, opened 2023-12-07
Fixed an issue with the `function.type` function where it might be unable to fetch the type information for a function.

arizvisa:
As reported and discussed in issue #175, the `function.type` function might actually fail at fetching the type from a database. This results in failures with regards to other functionality which depends on it. The original logic for `function.type` would first use the `idaapi.guess_tinfo` function to ask the disassembler to determine the type for a function prior to using the `idaapi.get_tinfo` function to get the actually applied type. Once both types have been fetched, the applied type will then be returned if it is available. If it is unavailable, then it falls back to returning the guessed type.

According to the report, it appears that there may be a condition where the disassembler may fail to both fetch type and guess it. Whenever this happens, a non-function type will be returned. Due to this non-function-type being returned, this caused functionality within the `database.functions` namespace to fail when trying to count the number of arguments for a function. This PR fixes the issue by fetching a type, guessing the type, and creating a default one in case those first two fail. This results in being able to guarantee an `idaapi.tinfo_t` is always returned with an `idaapi.func_type_data_t` attached so that things which depend on the type being a function will still be able to treat it as one.

This fixes issue #175.

### DIFF
--- a/base/function.py
+++ b/base/function.py
@@ -3202,16 +3202,37 @@ class type(object):
     def __new__(cls, func):
         '''Return the type information for the function `func` as an ``idaapi.tinfo_t``.'''
         _, ea = interface.addressOfRuntimeOrStatic(func)
+        fn = idaapi.get_func(ea)
 
-        # Guess the type information for the function ahead of time because
-        # they should _always_ have type information associated with them.
-        ti = idaapi.tinfo_t()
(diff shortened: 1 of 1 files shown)


arizvisa:
According to issue #175, the user is explicitly attaching their own methods to the `tinfo_t` which results in all boolean-related tests failing when trying to determine the size. This was a pain to troubleshoot, but that's how Python is.

Closing this PR since the issue was a mistake to begin with.

Issue #172, opened 2023-12-06
Wrong path on windows


Harding-Stardust:
The correct path is:
%APPDATA%\Hex-Rays\IDA Pro
Note the missing "roaming", this is included in %APPDATA%

https://github.com/arizvisa/ida-minsc/blob/af07f167a0992aed3f3a2a1670593880baa57820/README.md?plain=1#L39

arizvisa:
Both issue #172 and #173 will be combined in a future PR since they reside within the same file.

arizvisa:
This was closed by PR #190. If you feel this does not resolve your issue, please let me know and I can re-open.

Issue #173, opened 2023-12-06
Using double quotation marks makes this work on Windows also


Harding-Stardust:
```
pip install -r "requirements.txt"
```
https://github.com/arizvisa/ida-minsc/blob/af07f167a0992aed3f3a2a1670593880baa57820/README.md?plain=1#L58

arizvisa:
Both issue #172 and #173 will be combined in a future PR since they reside within the same file.

arizvisa:
This was closed by PR #190. If you feel this does not resolve your issue, please let me know and I can re-open.

PR #190, opened 2023-12-24
Added some minor fixes to the README.md file for users on the windows platform.

arizvisa:
As suggested in issues #172 and #173, some of the documentation that is suggested in the `README.md` file may not be clear for users on the Windows platform. This includes the command for using `pip` to install the requirements, which was using single-quotes instead of double-quotes, and an environment variable that was using a `$`-prefixed name instead of the `%`-cuddled one that is associated with the windows command prompt. If we wanted to use Powershell's syntax, we could use `$Env:APPDATA`, but since this is so minor that it should be implied...I really don't even think it's worth the effort. Nonetheless, this PR corrects both instances using the exact suggestions as requested by the user in their issue reports.

This fixes issue #172 and #173.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ to augment their reversing endeavors without distraction.
 
 Installation should be pretty simple and requires simply cloning the repository
 directly into the user's IDA user directory. On the Windows platform, this is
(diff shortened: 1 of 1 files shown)


arizvisa (commented):
The patch itself (and even the commit) is larger than the number of characters changed.

Issue #191, opened 2024-02-08
Inability to access the attributes of netnodes that use a tag of 0.


arizvisa:
The `internal.netnode` module provides wrappers that are used to access supvals, altvals, hashvals, and other information about a netnode.

Commit af65f9cd66c83098c7f9972d8149dd72eb027912 added support for explicitly specfiying the tag so that you can access the contents of netnodes like "$ vmm", "$ funcs", and "$ ignore micro". Due to the way that the tag is tested for defaultness, it is impossible to specify a tag of 0.

This makes it impossible to access certain pieces of the disassembler using these APIs.

arizvisa:
This was fixed by PR #192. The [@persistence-refactor](https://github.com/arizvisa/ida-minsc/tree/persistence-refactor) branch was also rebased and updated in response.

PR #192, opened 2024-02-08
Replaced the default parameter check for tags to use a ternary comparison instead of falseness.

arizvisa:
This PR fixes an issue with the `internal.netnode` module that is related to how it determines whether to use the default tag or the one that is provided by the user. Previously, the default tag was chosen by checking the tag parameter for falseness. This made it impossible to pass a tag of `0`, due to a zero-value being implicitly treated as false.

This issue was fixed by replacing all instances where the tag is being distinguished from the default parameter with an explicit identity comparison to `None`.

This fixes issue #191.


arizvisa (commented):
my eyes are glazing over...

Issue #199, opened 2026-05-25
The `internal.netnode.value.get` function does not allow you to fetch the value from a netnode as `bytes` or a `bytearray`.


arizvisa:
The implementation of `internal.netnode.value.get` uses the wrong API from the SDK when the "type" is specified as `bytes`. This would have the same effect as the `bytearray` type, but apparently the `bytearray` case was never actually implemented.

Some of the other functions, specifically ones that are used for removal of items from a netnode, return an inconsistent type on failure or success. The return type for these functions should be updated so that they are consistent.

arizvisa:
This is fixed by PR #200.

arizvisa:
PR #200 was just merged. Goodbye.

PR #200, opened 2026-05-25
Corrected some of the types used by functions within the `internal.netnode` module.

arizvisa:
Some of the functions from the `internal.netnode` module for removing the contents of a netnode return an inconsistent type when they should be returning a boolean. Specifically, the `remove` function incorrectly returns the void result from the API being used. The other type-related situation can be found in the `blob.remove` function which returns an integer. The primary issue, however, is that the `value` namespace doesn't actually support getting a netnode value as its raw data (`bytes`). This is due to the original implementation using the wrong API, `_ida_netnode.netnode_valstr`, instead of `_ida_netnode.netnode_valobj`.

This commit fixes each of the mentioned issues by adding support for handling an instance of an IDAPython `netnode` to `utils.get` which adds improved functionality to the `has` function. Both the `remove` and `blob.remove` functions were also rewritten to return a boolean based on checking their result. The `remove` function ensures that the netnode was actually removed, whereas `blob.remove` will track the size of the blob to determine whether it was successful or not.

Next, the implementation of `value.get` was fixed and reordered so that it could return its result as a `bytes` or a `bytearray` type. A couple more API functions were added to the `netnode` namespace, and a minor formatting issue in the `hash.repr` implementation was fixed.

This fixes issue #199.

### DIFF
--- a/base/_netnode.py
+++ b/base/_netnode.py
@@ -145,7 +145,12 @@ class utils(object):
     @classmethod
     def get(cls, index):
         '''Return the netnode for the provided `index`.'''
-        return netnode.get(index)
+        if isinstance(index, (six.integer_types, six.string_types)):
+            return netnode.get(index)
+
+        # in case we were given an instance of a netnode object.
+        realindex = netnode.index(index)
+        return netnode.get(realindex)
 
     @classmethod
     def range(cls):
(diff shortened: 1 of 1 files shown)


base/_netnode.py
```diff
@@ -467,7 +489,17 @@ def remove(cls, nodeidx, tag, start=0):
...
-        return netnode.delblob(node, start, tag)
+
+        # first grab the size so that we can verify that there is something to
+        # remove. if there's no size, then we can just return a failure.
+        size = netnode.blobsize(node, start, tag)
+        if not(size > 0):
+            return False
+
+        # now we can actually remove the blob and then verify that we actually
+        # removed something by checking the remved size with our preserved one.
```
arizvisa: type-o here. "remved" -> "removed".

arizvisa (commented):
some minor typing issues.

arizvisa (commented):
lookz good.

Issue #198, opened 2026-04-26
comment hook processing corrupts ARM autoanalysis / BL call flow


c0d3z3r0:
**Environment**

- IDA Pro 7.6
- Python 2
- ARM 32-bit firmware
- ida-minsc https://github.com/arizvisa/ida-minsc/commit/778f606737c13f5686a772f79fa643a4aebb3af5
- “Disable BL jumps detection” = true.

**Summary**

When ida-minsc is enabled during the initial autoanalysis of an ARM firmware, IDA incorrectly treats some normal BL instructions as jumps instead of calls. This causes functions to be split and/or not created correctly.

The issue disappears when ida-minsc is disabled. It also disappears when the changing_cmt / cmt_changed hooks are not registered, or more specifically when the `internal.comment.contents.inc(ea, key)` call is skipped.

This seems to be caused by ida-minsc updating its function-local comment/tag reference cache while IDA’s own autoanalysis is still building functions and code xrefs. Updating function-local metadata at this point seems to re-enter or disturb IDA’s analysis state. In this ARM firmware, the result is that IDA loses valid fallthrough xrefs after BL instructions and then fails to create valid functions.

This happens both during the initial analysis and also during a later manual reanalysis.

**Example**

With ida-minsc enabled, this ARM function is not recognized correctly:

```
.text:00325594                 MOV             R12, SP
.text:00325598                 PUSH            {R4,R5,R11,R12,LR,PC}
.text:0032559C                 SUB             R11, R12, #4
.text:003255A0                 MOV             R4, R0
.text:003255A4                 MOVW            R0, #0x1137
.text:003255A8                 BL              sub_2B2F14
.text:003255AC ; ---------------------------------------------------------------------------
.text:003255AC                 CMP             R0, #0
.text:003255B0                 BEQ             loc_325600
```

Without ida-minsc, IDA analyzes the same code correctly:

```
.text:00325594 ; =============== S U B R O U T I N E =======================================
.text:00325594 sub_325594
.text:00325594                 MOV             R12, SP
.text:00325598                 PUSH            {R4,R5,R11,R12,LR,PC}
.text:0032559C                 SUB             R11, R12, #4
.text:003255A0                 MOV             R4, R0
.text:003255A4                 MOVW            R0, #0x1137
.text:003255A8                 BL              sub_2B2F14
.text:003255AC                 CMP             R0, #0
...
.text:00325608                 LDMFD           SP, {R4,R5,R11,SP,PC}
[...]

PR #201, opened 2026-05-25
Usage of the tag cache interferes with processors that may use the same netnode for their implementation.

arizvisa:
The tag cache, written back in 2016, was incorrectly using the altvals for a function to store its index as a blob. This wasn't tested well against all of the supported architectures which resulted in breaking the disassemblers for both the AArch and AArch64 architectures. The main symptom that was discovered (as a result of this misdesign), was that the disassembler was not correctly identifying the return address from an instruction that belongs to a case (table-based switch).

This PR remedies the issue by moving the tag cache for a function into its own netnode that is distinctly separate from anything created by the disassembler. However, a lot more work had to be done due to backwards compatibility. So, the `contents` namespace was updated to add a number of classmethods for verifying where the tagcache for a function is located, and for reading and writing to that specific tagcache. The next addition is a classmethod for generating the unique netnode name for a function address so that the code can distinguish whether the tagcache was created or not.

Once these primitives were added, so wrappers were included in order to migrate the tagcache from its old location to the new one while creating the netnode if necessary. This then allowed modifying both the `contents._read` and `contents._write` classmethods so that the tagcache will be created or migrated on-demand.

This fixes issue #198.

### DIFF
--- a/base/_comment.py
+++ b/base/_comment.py
@@ -1231,9 +1231,12 @@ class contents(tagging):
                 if not ok:
                     logging.debug(u"{:s}._write({!r}, {:#x}, {!s}) : Unable to remove the address {:#x} from the cache header associated with the key {:#x}.".format('.'.join([__name__, cls.__name__]), target, ea, internal.utils.string.repr(value), ea, key))
 
+            # after removing the supvalue in the header for the given function,
+            # now we can delete the blob that is stored within the netnode for
+            # the function.
             finally:
-                count = internal.netnode.blob.remove(key, cls.btag)
-                logging.debug(u"{:s}._write({!r}, {:#x}, {!s}) : Removed {:d} blob{:s} ({!s}) associated with the key {:#x}.".format('.'.join([__name__, cls.__name__]), target, ea, internal.utils.string.repr(value), count, '' if count == 1 else 's', cls.btag, key))
+                ok = cls._del_tagcache_blob(key, ea)
+                logging.debug(u"{:s}._write({!r}, {:#x}, {!s}) : Removed the blob ({!s}) associated with the key {:#x}.".format('.'.join([__name__, cls.__name__]), target, ea, internal.utils.string.repr(value), cls.btag, key))
 
             return True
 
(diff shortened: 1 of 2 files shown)


arizvisa:
I suspect there's an issue when a function entrypoint is modified that can result in losing the tagcache for a function. This is because the way that we generate the name from a function's address.

Before implementing this fix entirely within the implementation of the tagcache, I was originally considering using the IDB hooks (`ui.hooks.idb`) to track when a function was edited (`func_added`, `func_updated`, or `func_deleted`) so that its netnode could be removed. However, it might be better to use the `set_func_start` event instead. These events should probably be tested with and verified once it is confirmed that this bug actually exists.

arizvisa:
The first iteration of this PR had an issue with it due to the locator-type functions (`contents._has_old_tagcache`, `contents._nas_new_tagcache`, and `contents._migrate_tagcache`) auto-correcting the address that it was given so that it points at the function entrypoint. Due to the correction, the implementation couldn't reliably check for the existence of a tagcache for the specified address and would cause multiple exceptions to be raised when building the testcase.

The implementation of these functions were modified (during the previous force-push) to remove the function entrypoint auto-correction which requires each to use the \_exact\_ address. This modification allows these functions to be used accurately, and to only migrate the tagcache if the exact address is used.

There is still an issue when the address of a function is changed that results in the netnode containing the tag cache for a specific function to have the wrong name. From testing, it seems that the disassembler recreates the cache using the new address anyways. However, this means that the original netnode containing the tag cache has been orphaned.

As a workaround, I'm considering adding another hook that will remove the old netnode if the function entrypoint is changed. This is because relocating the data by copying it to the new location might affect any new changes created by the disassembler as a result of modifying the function entrypoint.

arizvisa:
Modified the `set_func_start` hook so that it attempts to rename the netnode containing the old tagcache to the one for the new function address. If that fails, it will copy the contents from the old tagcache into the new one before removing the old netnode for it.

arizvisa:
Just waiting on confirmation from the original reporter.

